PR #2, opened 2015-05-04
Add support for WordPress Multisite installations.

aaronhuisinga:
The current version of the plugin appears in the Tools menu of every
site within a WordPress Multisite installation. In the case of a
Multisite installation, the administration options should only be
available to users with the Manage Network Options capability, and the
settings page should show within the Network section of the admin panel.

Note: I've also updated all calls to the admin_url function to network_admin_url. This is necessary for the plugin to work on WordPress Multisite installations, and simply reverts to the expected admin_url function if the installation is not multisite.

### DIFF
--- a/includes/admin-page.php
+++ b/includes/admin-page.php
@@ -76,14 +76,14 @@
 	<p class="submit">
 
 		<?php if ( strcasecmp( 'connected', $this->get_redis_status() ) === 0 ) : ?>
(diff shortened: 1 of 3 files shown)


aaronhuisinga:
I also went ahead and bumped the version within the files to v1.2. If you have something else in mind for your versioning, just update those before merging and updating releases.

tillkruss:
Thanks Aaron!

aaronhuisinga:
Not a problem! Let me know when you get the plugin updated in the WordPress repository so I can use the official distro instead of my customized one.

PR #5, opened 2015-06-23
Fixed WP_CACHE_KEY_SALT

zorrobyte:
Uses md5(DB_NAME . __FILE__) to actually be unique, was just set to an empty string before..

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -449,7 +449,7 @@ class WP_Object_Cache {
 		 * multi single WP installs on the same server.
 		 */
 		if ( ! defined( 'WP_CACHE_KEY_SALT' ) ) {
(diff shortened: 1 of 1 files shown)


tillkruss:
Key are already automatically prefixed with the WordPress `$table_prefix`.

zorrobyte:
False, this doesn't work as explained and caused collisions between WordPress versions. Check your code.

zorrobyte:
The issue is that with cPanel/WHM installations, that the $table_prefix is the same across installations on the same account causing collisions. If another method was used to guarantee uniqueness, you may see better adoption of your plugin as not every user wants, needs or knows how to setup a table prefix (and server admins won't be annoyed in managing tons of installs).

tillkruss:
Yeah, if more users run into the same issue, I'm going to implement a solution.

PR #15, opened 2015-10-28
Use WordPress constant to load Predis library

carlalexander:
Modified object cache to use `WP_CONTENT_DIR` instead `dirname( __FILE__)`. This fixes an issue where the object cache couldn't find the Predis library. The cause was the fact that `object-cache.php` was symlinked into `wp-content`. `dirname( __FILE__)` would resolve to the wrong location.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -419,11 +419,11 @@ class WP_Object_Cache {
 				}
 
 				// check if bundled Predis library exists
-				if ( ! realpath( dirname( __FILE__ ) . '/plugins/redis-cache/includes/predis.php' ) ) {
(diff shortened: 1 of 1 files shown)


carlalexander:
Just as note, I also tested `WP_PLUGIN_DIR`. The issue was that the constant isn't defined yet when WordPress loads `object-cache.php`. Because of that, we have to use `WP_CONTENT_DIR`.

jeremyclarke:
+1 This is vital for our usage.

tillkruss:
Would it make sense to check `dirname( __FILE__ )` as fallback as well, just in case someone has an even more complicated setup?

carlalexander:
I can't think of another edge case right now that isn't covered by the `WP_CONTENT_DIR` constant.

PR #22, opened 2016-01-28
Fixing crash that occurs when connection fails

scott-kennedy:
There are currently three possible Redis clients that handle connections in different
ways and this has lead to a disconnect with how connections are handled.  This pull
request adds a bit of consistency when connecting to the Redis server.

To reproduce this issue you'll need to set your `WP_REDIS_CLIENT` to pecl and then
run your application without starting your Redis server. (this is an easy way to
guarantee a failed connection)

Previously only the hhvm client would check the success of the connect call while the
other two clients just assumed a connection occurred.

Proposed solution is to remove the explicit checking from the hhvm client so it's
inline with the behavior of the other two clients.  After we've set up our connection
but before we leave the try block we'll attempt to ping the server which will fail if
a proper connection wasn't established giving us the desired behavior of not setting
`$this->redis_connected`.

Please let me know if I can provide more information or you'd like me to change my approach.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -374,9 +374,7 @@ class WP_Object_Cache {
 					$redis[ 'port' ] = 0;
 				}
 
-				if ( ! $this->redis->connect( $redis[ 'host' ], $redis[ 'port' ] ) ) {
-					throw new Exception;
-				}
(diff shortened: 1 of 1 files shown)


tillkruss:
I'm going offline for a couple of days, I'll check this out next week.

PR #27, opened 2016-04-09
Added support for $found parameter

KristianI:
Added support for $found parameter according to
https://wordpress.org/support/topic/boolean-values


tillkruss:
Thanks, can also actually make use of `$force` in the method?

PR #29, opened 2016-04-30
Check core functions exist

ethitter:
The object cache can be used inside of an `advanced-cache.php` dropin, but that also means that the object cache can be loaded early enough that most of Core hasn't been included yet, notably the actions and filters callbacks.

Checking that Core functions exist first avoids fatal errors with plugins like Batcache, or any other `advanced-cache.php` dropin that leverages the object cache.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1,4 +1,4 @@
-®<?php
+<?php
 /*
 Plugin Name: Redis Object Cache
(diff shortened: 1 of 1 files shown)


pnu:
👍 We have our own fork with the exact same fix, for the exact same reason. Please merge.

tillkruss:
Thanks!

PR #32, opened 2016-06-28
update ::delete() with redis_object_cache_delete action

jeremyclarke:
This is similar to the  get/set actions (and get filter) already added in a previous update. 
Just adds a matching action for delete. 

Personally I need this for my logging system that tracks actions and specifically because my 'set' action triggers delete commands and I need them logged to. 

Don't think there's any real risk to adding this and it would be very useful for me. 

Thanks!


jeremyclarke:
Danke!

tillkruss:
Bitte : )

PR #33, opened 2016-07-20
Flush per site cache

Sneezry:
Redis Object Cache provides a Flush Cache button, however, it clears the entire DB. We have a multisite WordPress with thousands of blogs, and to flush all cache may lead performance problem to us. So I make some changes to flush per site cache. This feature is only available for network sites. A Flush Cache link will be added in each site line in Network Sites menu.

flush_cache_action

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -94,6 +94,20 @@ function wp_cache_flush( $delay = 0 ) {
 }
 
 /**
+ * Invalidate all items in the branch.
+ *
+ * @param int $branch  The name of the branch will be deleted.
+ *
+ * @global WP_Object_Cache $wp_object_cache
+ *
+ * @return bool             Returns TRUE on success or FALSE on failure.
+ */
+function wp_cache_delete_branch( $branch ) {
+	global $wp_object_cache;
+	return $wp_object_cache->delete_branch( $branch );
+}
+
+/**
  * Retrieve object from cache.
  *
  * Gets an object from cache based on $key and $group.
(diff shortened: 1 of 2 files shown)


tillkruss:
How long have you been using this in production already?

Sneezry:
We haven't used this in production, because if we update the plugin the feature will be lost, and that's the reason I create this pull request. However, the code has passed our code review :)

tillkruss:
I'd be open to merging this after you tested this feature thoroughly. Any chance you guys could manually apply the changes and see how it goes?

Sneezry:
OK, we will test it on our dev environment for days, and if it works fine, we will deploy it on MSDN blog production. I will inform you when we decide to deploy it on production, and you may consider when to merge it then :)

Sneezry:
The plugin has been tested on PPE (test env) for a half month, and we have deployed it on our production today. I move "Flush Cache" link from custom column to row action as shown in our production below:

download

tillkruss:
Nice, keep me posted.

lkraav:
This is terrific, I'm super interested.

tillkruss:
@Sneezry: Can you fix the conflicts?

Sneezry:
OK, I'll fix it in few days.

tillkruss:
This would be nice to have as well: https://make.wordpress.org/core/2016/10/04/custom-bulk-actions/

tillkruss:
https://github.com/tillkruss/redis-cache/pull/53

tillkruss:
We just implemented selective cache key flushing that uses the `WP_CACHE_KEY_SALT` constant in https://github.com/tillkruss/redis-cache/pull/53 and I have to say I'm not sure if a per-site cache flush makes sense to me.

It seems like a cache-invalidation nightmare to me because of the global groups, like the `users` table. If someone can describe a scenario in which they would greatly benefit from a per-site flush and they have a solid solution for this let me know. Otherwise prefer not to merge this.

Sneezry:
Woooow! That is great! My code has to cache all key name in an index, and
that makes redis write twice. Code in #53 is awesome!

Till Krüss <notifications@github.com>于2016年11月28日 周一上午7:09写道：

svandragt:
So are you saying you're thinking of moving away from this PR in favour of a cache salt approach (https://github.com/tillkruss/redis-cache/pull/53)? In a multisite scenario then would one have to all the site-id to the cache salt to have functionality similar to this?

Being able to clear a cache for a single site in a high traffic multisite environment is a definite use case that we have as flushing the db causes a large performance drop. Typically in this environment visitors  do not login with WordPress therefore not flushing the users table is ok.

However I can also see implementations where this could be a problem, so perhaps this PR simple needs a setting/option for flushing the 'shared cache' along with a site flush so that we can  have that control.

tillkruss:
@svandragt: No, I'd still like to see better multisite support in the plugin, but it has to be implemented in a way that works great out-of-the-box for "regular" users and we can "unlock" different scenarios with `WP_REDIS_*` constants.

A few thoughts:

By default the entire cache should be flushed, multisite or not.

Next level could be a constant that enables per-site flushing, including all shared/global groups via actions under "Sites". However it must still be possible to flush the entire cache somewhere (maybe _Network -> Settings -> Redis_).

There should be WP CLI commands for flushing:
- the entire cache
- the cache of a single network site
- the global groups

If another separate constant is added that prevents the global groups from being flushed in multisite environments, then there should be another button for that, maybe as well under _Network -> Settings -> Redis_.

ravloony:
> If someone can describe a scenario in which they would greatly benefit from a per-site flush and they have a solid solution for this let me know. Otherwise prefer not to merge this.

We are using a multisite setup (~4000 websites and counting) where we each site has an internal url of the form slug.ourcompany.com until the site moves to production, at which point our provisioner application edits all urls in the database for that site to change them to the real url. This implies a cache flush.

Obviously, with that many sites, a cache flush is be pretty damaging to performance. As it stands we are holding of our impending doom by cunningly queueing connections in nginx so the database never gets overwhelmed, but ideally we would like a per site flush to be possible.

One option I am toying with is to set the WP_CACHE_KEY_SALT to md5($_SERVER['SERVER_NAME']) and using WP_REDIS_SELECTIVE_FLUSH, but I am not yet sure of the implications. For one thing it may make it difficult to use wp-cli.

tillkruss:
@ravloony: That should work fine, better than flushing all sites at once. Depends how many hits you have per second on those 4000 sites.

I never use hashes as prefixes, because it makes impossible to find manually, instead I usually use a sanitized string of the domain, like `example_com:` or `beta_example_com:`.

tillkruss:
To be honest, I don't like the approach of this PR. Using the prefix for `WP_REDIS_SELECTIVE_FLUSH` seems better to me.

The "Flush Cache" UI link is neat. We should use the selective flush constant tho and allow cache flushing via network admin and normal admin UI. If anyone wants to make a PR, that'd be great!
I think it would be nicer if the a "Flush Site"

PR #47, opened 2016-10-04
Allow disabling of the object cache at runtime.

rkaiser0324:
Even if the plugin is activated and enabled, now you can disable it via a define in wp-config.php.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,10 @@ For detailed installation instructions, please read the [standard installation p
 
 1. Make sure [Redis is installed and running](http://redis.io/topics/quickstart).
 2. Install and activate plugin.
-3. Enable the object cache under _Settings -> Redis_.
+3. Enable the object cache under _Settings -> Redis_.  After doing so, you can optionally toggle it off at runtime via adding this to `wp-config.php`:
+```php
+define('REDIS_OBJECT_CACHE_DISABLE', true);
+```
 4. If necessary, adjust [connection parameters](http://wordpress.org/extend/plugins/redis-cache/other_notes/).
 
 If your server doesn't support the [WordPress Filesystem API](https://codex.wordpress.org/Filesystem_API), you have to manually copy the `object-cache.php` file from the `/plugins/redis-cache/includes/` directory to the `/wp-content/` directory.
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1,1110 +1,1114 @@
 <?php
 /*
-  Plugin Name: Redis Object Cache
-  Plugin URI: http://wordpress.org/plugins/redis-cache/
-  Description: A persistent object cache backend powered by Redis. Supports Predis, PhpRedis, HHVM, replication, clustering and WP-CLI.
-  Version: 1.3.4
-  Author: Till KrÃ¼ss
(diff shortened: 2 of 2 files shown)


tillkruss:
Can you fix the whitespace?

rkaiser0324:
OK, I reverted to your original whitespaces as that should work better.  Not sure why the StyleCI analysis is still failing though, as the diff only changed 5 lines?

tillkruss:
If you don't mind to avoid the encoding issues, I'll close this and implement this slightly different.

rkaiser0324:
Be my guest.  I was surprised I couldn't just "return" and had to do a conditional here.  Thanks.

tillkruss:
Could you explain your use case for this? And why you can't call `wp redis disable` instead?

rkaiser0324:
Sure.  We deploy to our production cluster via "git pull", and want our development DB and codebase to match production exactly (except for wp-config.php).  When you activate and enable this plugin it creates wp-content/object-cache.php - so we source control that as well.

In development we typically want to bypass object caching, but don't want to delete wp-content/object-cache.php (which dirties the working tree) or deactivate the plugin (which dirties the DB).  So we would like to leave both as-is, and simply disable caching via a directive in wp-config.php.

As you suggested I looked through WP core to see if there was another way to accomplish this via filters etc., but didn't find one - so ended up with the rather ugly conditional you see here.

tillkruss:
Okay, thanks. Do you use any deployment scripts, like Composer, Capistrano or similar? Because in that case you can just run `wp redis enable` after the deployment to enable dropin and you don't have to include the `object-cache.php` in your repository.

I digged into this last night and the only solution seems to be wrapping the entire file in a condition and I really don't like that idea.

rkaiser0324:
We use Composer for some stuff, but the majority is "git pull".  I get what you're saying though.

No worries, I'll just use my fork.  Thanks for your work on this plugin, I can't imagine trying to run WordPress without it.

tillkruss:
The constant `WP_REDIS_DISABLED` is now available : )

jeremyclarke:
Thanks! Great ti know it's there!

fagiani:
@tillkruss thanks for the great work on this plugin!

Is there a way to force the plugin to be enabled via constant? Would `WP_REDIS_DISABLED=false` ensure that? I ask that because of a common scenario of many users with a setup of multiple application servers behind a load balancer, it looks like at the moment you need to enable the plugin on each app server manually (at least the button on the plugin page indicates that).

Also it would be interesting to check weather the plugin is installed as must use (on `mu-plugins`) before replacing the `wp-content/object-cache.php`. Must use plugins are very important when you want to make sure advanced users won't be able to disable the plugin or remove them.

Keep Rocking!

tillkruss:
IIRC `object-cache.php` is loaded long before `mu-plugins`.

fagiani:
@tillkruss I was able to change the `object-cache.php` permissions to `400` and on the line where it calls `plugins` on the path i changed to `mu-plugins` and that worked great as it won't get overridden and this way all application servers got redis enabled by default. Hopefully that helps other folks in the future.

Thanks!

Issue #52, opened 2016-11-18
respect key salt when purging?


pjv:
i don't know about anyone else, but i was kind of assuming that having defined a specific salt for a given WP site that when i used the "flush cache" button in the redis settings for that WP site that i would only be flushing cache entries that matched that key salt.

for a server that has more than one WP site, and if you are also using redis for page caching like i am, hitting that flush cache button can have a pretty big unintended effect. i built my own command line flush using redis-cli that lets me be a little more precise about what i am flushing, but i still think it would be cool to have the flush cache button respect the key salt if possible.

having done some poking around, i see why it is the way it is. telling redis to flushdb is fast, easy, atomic and easy to implement with any back end to redis. there are various hacks around for doing wildcard purging on redis. my command line purger uses one of them based on redis `SCAN` which works fine for my purposes but it is not guaranteed atomic.

what seems like the best approach if you want an atomic wildcard delete is using a small lua script against the redis db (something like http://blog.code4hire.com/2016/09/Deleting-keys-with-wildcards-in-Redis/).

is this something you'd be interested in incorporating into the plugin? if so, when i get some time i'd be willing to try adding it to the existing flush method in `object-cache.php` and sending a PR if i get it working. if not, i'm ok with my command-line tools.

tillkruss:
Yes, that would be nice, send a PR! Checkout https://github.com/tillkruss/redis-cache/pull/33/files as well.

The best practise right now is to use a separate Redis database for each installation and this should remain the the default for now.

Your alternative delete method should be:
- enabled by a constant
- and only be effective in multisite environments
- or when a cache key prefix is set

pjv:
i don't use WP multi-site - though i do have multiple separate WP installs per server. and i did see that existing multi-site PR. assuming that gets merged at some point, and given that i don't use multi-site, i was envisioning my patch as filling a similar role but for non-multi-site.

i was thinking that it would be enabled by having set the cache key prefix constant. but maybe that's too big a change in functionality to foist on existing users.

i was really thinking about it in big crayon terms something like, if we are marking everything we put into redis with our prefix, then when we are flushing we should only flush what we put in there.

tillkruss:
Yeah, definitively introduce another constant aside from `WP_CACHE_KEY_SALT`, for backwards compatibility. I'd also rather see your LUA script than the other PR to be honest, if you don't add Multisite support, I will after it's merged.

pjv:
[...]

PR #53, opened 2016-11-27
Flush cache with key prefix

pjv:
See discussion in #52.

This patch seems to be working well on my production box. I think it's good to go for my use-case. I haven't tested it with multi-site and I haven't tested it with YUUGE numbers of keys, but I have tested it with 400K+ keys and it took < 1 second to flush them.

Yay lua.


tillkruss:
Thanks! I've adjusted your code a little bit in https://github.com/tillkruss/redis-cache/commit/4b8ba3852877bcb2c8001e478b85e0e453ada42a.

pjv:
cool, your code is much purtier than mine. i'll adjust my wp-config defines and push your version to my boxes and look forward to the next plugin update.

PR #57, opened 2017-02-16
Add sharding support for pecl redis (phpredis)

ThatGerber:


tillkruss:
Would it make more sense to use `RedisCluster`?

ThatGerber:
That would be for clusters (replication https://github.com/phpredis/phpredis/blob/develop/cluster.markdown). This example is to support an array of the databases (sharding https://github.com/phpredis/phpredis/blob/develop/arrays.markdown).

tillkruss:
Yeah. What I mean is that `WP_REDIS_CLUSTER` creates a cluster using Predis, not sharing. I think we should add a `WP_REDIS_SHARDS` constant and extend the Predis connector as well. If you're cool with that, I'll merge and add rename the constant.

tillkruss:
Okay, checkout https://github.com/tillkruss/redis-cache/commit/42194abf7c04497f3e3cb2fd150827344af57e6c.

I probably won't release a stable version of this for a few weeks until a couple of people tested this.

ThatGerber:
Sorry, reading through the cluster spec thoroughly to make sure that everything is labeled correctly. Looks like there's a few places where "sharding" is spelled "sharing". I'll send a PR with some fixes.

I'm curious whether it's useful to define another constant. The three boil down to basically a constant for establishing what the connection strings are for an array of Redis nodes. Would it be clear to use the constant as a way of declaring connection strings, and then create another constant that determines the partitioning method (https://redis.io/topics/partitioning#different-implementations-of-partitioning). The 3.0 cluster is a mix of client-side sharding and query routing done on the nodes. 

An example configuration:

```php
define('WP_REDIS_CLUSTER_TYPE', 'sharding');

define( 'WP_REDIS_NODES', [
    'tcp://127.0.0.1:6379?database=10&alias=Node-01',
    'tcp://127.0.0.2:6379?database=10&alias=Node-02',
    'tcp://127.0.0.3:6379?database=10&alias=Node-03',
] );
```

tillkruss:
Thanks, I fixed the typos.

I like your idea, even tho it would break existing functionality. Checkout https://github.com/tillkruss/redis-cache/pull/56, that could be tackled as well with the same approach.

PR #59, opened 2017-03-06
Force expiration time for keys higher than zero exporation time

aytackokus:
This solution did solve our problem with creation of 100000+ keys after few hours of caching.

That happend because the wordpress does cache the terms for a day and you cannot change that as it is in the core.

Currently we have around 10K keys and the website speed is very fast.

FYI it does keep the default expiration value if you don't define the custom TTL.


tillkruss:
Thanks! I've changed the logic a little bit to always enforce the `WP_REDIS_MAXTTL` if it's set: https://github.com/tillkruss/redis-cache/commit/ba27d2369c5cfafb9b57df8cd894860c074845dd

PR #66, opened 2018-03-12
Use the PHAR version of Predis

synotio:
Fixes https://github.com/tillkruss/redis-cache/issues/65

It seems that there are bugs in the "create-single-file" script for predis. I decided to use create-phar instead which seems to work perfectly.

The name is still predis.php in order to maintain backwards compatibility. Older versions of object-cache.php will still require() redis-cache/includes/predis.php and so the name must not be changed (maybe change object-cache.php to just a stub loader for future releases so that it can be updated when the plugin is updated?).


tillkruss:
Cool. And that just works?

synotio:
Yep, works perfectly!

tillkruss:
I'm reverting this because of #67. If there is something wrong with the `create-single-file` script, please send a PR to https://github.com/nrk/predis.

PR #70, opened 2018-04-04
Fix RedisCluster warnings

boboudreau:
A warning is generated when using a `RedisCluster` as the datasource for object cache with PECL Redis:
```
PHP Warning:  RedisCluster::ping() expects exactly 1 parameter, 0 given in /var/www/sportshub/releases/20180402235221/web/app/object-cache.php on line 494
```

This PR updates object-cache.php to ping with a host when in cluster mode. This avoids continuous PHP warnings in logs. I chose `WP_REDIS_HOST` as the constant to use since it's required in single node implementations and makes sense if you're using a hosting provider that only requires you know the endpoint to a single server to make use of an entire Redis Cluster, like AWS: https://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/Endpoints.html#Endpoints.Find.RedisCluster

Of course we could use a different constant, or a random entry from the seed nodes passed in as well, since it's just checking to see if the connection is alive.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -424,9 +424,9 @@ class WP_Object_Cache
                 $this->redis_client = sprintf('PhpRedis (v%s)', phpversion('redis'));
 
                 if (defined('WP_REDIS_SHARDS')) {
-                    $this->redis = new RedisArray(explode(',', WP_REDIS_CLUSTER));
+                    $this->redis = new RedisArray(WP_REDIS_CLUSTER);
                 } elseif (defined('WP_REDIS_CLUSTER')) {
-                    $this->redis = new RedisCluster(null, explode(',', WP_REDIS_CLUSTER));
+                    $this->redis = new RedisCluster(null, WP_REDIS_CLUSTER);
                 } else {
                     $this->redis = new Redis();
 
@@ -474,7 +474,7 @@ class WP_Object_Cache
                     $parameters = WP_REDIS_SERVERS;
                     $options[ 'replication' ] = true;
                 } elseif (defined('WP_REDIS_CLUSTER')) {
-                    $parameters = explode(',', WP_REDIS_CLUSTER);
+                    $parameters = WP_REDIS_CLUSTER;
                     $options[ 'cluster' ] = 'redis';
                 }
 
@@ -683,7 +683,13 @@ class WP_Object_Cache
                     $this->redis instanceof Predis\Client ? 0 : []
                 ));
             } else {
-                $result = $this->parse_redis_response($this->redis->flushdb());
+                if (defined('WP_REDIS_CLUSTER') && defined('WP_REDIS_CLIENT') && strcasecmp('pecl', WP_REDIS_CLIENT) === 0) {
+                    foreach(WP_REDIS_CLUSTER as $host) {
+                        $result = $this->parse_redis_response($this->redis->flushdb($host));
+                    }
+                } else {
+                    $result = $this->parse_redis_response($this->redis->flushdb());
+                }
             }
 
             if (function_exists('do_action')) {
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -491,7 +491,11 @@ class WP_Object_Cache
             }
 
             // Throws exception if Redis is unavailable
-            $this->redis->ping();
+            if (defined('WP_REDIS_CLUSTER') && strcasecmp('pecl', $client) === 0) {
+                $this->redis->ping(WP_REDIS_HOST);
+            } else {
+                $this->redis->ping();
+            }
 
             $this->redis_connected = true;
         } catch (Exception $exception) {
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -424,9 +424,9 @@ class WP_Object_Cache
                 $this->redis_client = sprintf('PhpRedis (v%s)', phpversion('redis'));
 
                 if (defined('WP_REDIS_SHARDS')) {
-                    $this->redis = new RedisArray(WP_REDIS_CLUSTER);
+                    $this->redis = new RedisArray(array_values(WP_REDIS_CLUSTER));
                 } elseif (defined('WP_REDIS_CLUSTER')) {
-                    $this->redis = new RedisCluster(null, WP_REDIS_CLUSTER);
+                    $this->redis = new RedisCluster(null, array_values(WP_REDIS_CLUSTER));
                 } else {
                     $this->redis = new Redis();
 
@@ -492,7 +492,7 @@ class WP_Object_Cache
 
             // Throws exception if Redis is unavailable
             if (defined('WP_REDIS_CLUSTER') && strcasecmp('pecl', $client) === 0) {
-                $this->redis->ping(WP_REDIS_HOST);
+                $this->redis->ping(current(array_values(WP_REDIS_CLUSTER)));
             } else {
                 $this->redis->ping();
             }
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -474,7 +474,7 @@ class WP_Object_Cache
                     $parameters = WP_REDIS_SERVERS;
                     $options[ 'replication' ] = true;
                 } elseif (defined('WP_REDIS_CLUSTER')) {
-                    $parameters = explode(',' WP_REDIS_CLUSTER);
+                    $parameters = explode(',', WP_REDIS_CLUSTER);
                     $options[ 'cluster' ] = 'redis';
                 }
 
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -661,7 +661,7 @@ class WP_Object_Cache
             sleep($delay);
         }
 
-        $result = false;
+        $result = [];
         $this->cache = array();
 
         if ($this->redis_status()) {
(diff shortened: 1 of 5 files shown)


boboudreau:
Sure - when I ran this locally `Redis` was defined so it didn't get into the `Predis` block of code, and I only added some tests here to get you started.  There's a lot of work to do if we wanted to build adapters for `Redis`, `RedisCluster`, `RedisArray`, etc., in order to be able to mock them and run more interesting tests.

I'll need to play around with Travis to figure out how to get it to use PECL Redis.  Unless you prefer it using `Predis`, in which case, there's going to be even more hacking involved in Travis because https://github.com/tillkruss/redis-cache/blob/master/includes/object-cache.php#L461-L462 kind of hard-codes the location of the library.

I'll see what I can do in my own fork (I believe I set up Travis properly) before re-opening this PR.

boboudreau:
You'll see that there was an additional change in handling the `WP_REDIS_CLUSTER` constant here. PHP 5.6 does not allow array values in a constant, and now that we're actually running some tests in Travis, it was caught. 👍 

So, defining a cluster now needs to be done as a comma separated list. Anyone defining their cluster as a single endpoint (as AWS requires) will not be affected by this change.

includes/object-cache.php
```diff
@@ -474,7 +474,7 @@ public function __construct($fail_gracefully = true)
                     $parameters = WP_REDIS_SERVERS;
                     $options[ 'replication' ] = true;
                 } elseif (defined('WP_REDIS_CLUSTER')) {
-                    $parameters = WP_REDIS_CLUSTER;
+                    $parameters = explode(',', WP_REDIS_CLUSTER);
```
tillkruss: Can you explain those 3 `explode()` calls to me? Why are you switching?

boboudreau:
Looks like `flushdb()` is also affected.  It takes 0 arguments for `Redis()` and `RedisArray()`, but 1 (or 2) arguments for `RedisCluster()`:

https://github.com/phpredis/phpredis/blob/master/redis.c#L1735
https://github.com/phpredis/phpredis/blob/master/redis_array.c#L670
https://github.com/phpredis/phpredis/blob/master/redis_cluster.c#L2616

I'll be updating the PR to account for RedisCluster::flushdb() here: https://github.com/tillkruss/redis-cache/blob/master/includes/object-cache.php#L682

Let me know what you think you want to do re: the constants and PHP 5.6 compatibility.

boboudreau:
I'm going to revert my `explode` code since you've documented that these constants must be arrays, even though that doesn't work in PHP 5.6. 

In the meantime, you can either drop 5.6 from the travis tests and update your package to say minimum PHP version is 7.0, document that this library does NOT support Redis Clustering w/ PECL Redis on PHP 5.6, or you'd need to change versioning and introduce a backward incompatible change. Feel free to update my branch so it aligns with what you want. I'll mark the RedisCluster test as skipped for  PHP 5.6 testing.

includes/object-cache.php
tillkruss: Let's stick with PHP 7 only. 5.6 is EOL anyhow.

tillkruss:
Yeah, feel free to update the Travis config.

boboudreau:
OK, this appears to be working on my PECL Redis & cluster environment. No more commits 😄

boboudreau:
I lied. I don't know if this is an AWS issue, or a RedisCluster issue, since I don't have any other systems to test on.  Basically, connecting to the server requires a `host:port` string, but flushing the cache on certain nodes requires a `host` string only - no port.  I'll update the PR to handle this - it'd be interesting to see if anyone with any other configurations sees similar behavior.

includes/object-cache.php
```diff
@@ -679,7 +683,14 @@ public function flush($delay = 0)
                     $this->redis instanceof Predis\Client ? 0 : []
                 ));
             } else {
-                $result = $this->parse_redis_response($this->redis->flushdb());
+                if (defined('WP_REDIS_CLUSTER') && defined('WP_REDIS_CLIENT') && strcasecmp('pecl', WP_REDIS_CLIENT) === 0) {
+                    foreach(WP_REDIS_CLUSTER as $host) {
+                        $parsed_host = parse_url($host);
+                        $result = $this->parse_redis_response($this->redis->flushdb($parsed_host['host']));
+                    }
```
tillkruss: I never use Redis Cluster. Is it necessary to call `FLUSHDB` on all instances, wouldn't one be enough?

boboudreau: Correct. The documentation for PHPRedis claims that FLUSHDB is a single-node interaction, and must be called on a specific node: https://github.com/phpredis/phpredis/blob/develop/cluster.markdown#directed-node-commands.

Based on the example code there, though, I may try seeing if I can use the suggested syntax:

```
foreach($this-redis->_masters() as $master) {
    $this->redis->flushdb($master);
}
```

Will report back if that seems to work, and update my PR accordingly. This is all new to me, too, so apologies for all the different versions trying to get something to work.

boboudreau:
Just so this comment isn't lost in the response to outdated code:

> One thing that does NOT work is running LUA scripts, to, for example, delete all keys for a single site with a given prefix. In this object cache implementation, the derived key generated is:
> 
> {$salt}{$prefix}:{$group}:{$key}
> Because the prefix is not a hash tag, keys for items in the same blog are split across nodes, and cannot easily be deleted with a script. Would you consider a PR which updates the keys to use hash tags? You requested a similar feature for EVAL in the Predis implementation: nrk/predis#405 (comment)
> 
> I believe we could do something as simple as:
> 
> return "{{$salt}{$prefix}}:{$group}:{$key}";
> If you'd like, I can include that in another PR. If not, I'll probably need to use a fork of the object cache, since clearing the object cache for a single site in a multisite install using Redis Cluster will be important to me going forward.
> 
> I've just verified it works - though with a quirk. You need to iterate over all masters in the cluster and run the script on each node - or, even better, handle the MOVED response from the server to know which node to run the script on.

includes/object-cache.php
```diff
@@ -679,7 +683,13 @@ public function flush($delay = 0)
                     $this->redis instanceof Predis\Client ? 0 : []
                 ));
             } else {
-                $result = $this->parse_redis_response($this->redis->flushdb());
+                if (defined('WP_REDIS_CLUSTER') && defined('WP_REDIS_CLIENT') && strcasecmp('pecl', WP_REDIS_CLIENT) === 0) {
+                    foreach($this->redis->_masters() as $master) {
+                        $result = $this->parse_redis_response($this->redis->flushdb($master));
+                    }
```
tillkruss: I'd probably collect the `$results` and pass them to the `redis_object_cache_flush` action.

tillkruss:
@boboudreau: I'd either add a call to `error_log()` instead of calling `eval()`, or feel free to add support, that'd would be great of course!

boboudreau:
@tillkruss What would you like to log? 

Did you have any thoughts on my suggestion to group blog keys by using hash tags and changing the derived key? I've updated the PR to make the change to the derived key and fix the implementation to attempt to run LUA scripts on all nodes.  Without changing the derived key, `{salt}{blog_id}:` prefixed objects could land in different nodes, and LUA scripts would fail. 

Since there's no way to know which server to run the script on, this PR runs the script on all master nodes.

includes/object-cache.php
```diff
@@ -964,7 +991,7 @@ public function build_key($key, $group = 'default')
         $salt = defined('WP_CACHE_KEY_SALT') ? trim(WP_CACHE_KEY_SALT) : '';
         $prefix = in_array($group, $this->global_groups) ? $this->global_prefix : $this->blog_prefix;
 
-        return "{$salt}{$prefix}:{$group}:{$key}";
+        return "{{$salt}{$prefix}}:{$group}:{$key}";
```
tillkruss: Could you explain this nested tag?

tillkruss:
Log whatever makes sense to you, I might cleanup the the message after merging.

It all looks really good to me. I'll probably release this on a branch for people to test a while on production, simply to avoid what #66 did.

tillkruss:
@boboudreau: Have you been using this in production? Any issues?

pySilver:
any news?

tillkruss:
I merged this into the `cluster-fix` branch for now.

boboudreau:
@tillkruss @pySilver We've had this running in production since April with no issues. The `{{$salt}{$prefix}}:{$group}:{$key}` change in particular was extremely helpful grouping all multisite blog keys onto a single cluster, making it much easier to purge on a per-site basis.

tillkruss:
I've merged this into `master` now. Will go live with next release.

boboudreau:
@tillkruss that's fantastic. I confirmed from inspecting your master branch (where the object cache is listed as version 1.4.2) that it includes all of the changes required to get an AWS ElastiCache Redis Cluster working for our purposes on a multi-site network install of WordPress.

Any thoughts on when your next release might be? Or how stable you feel the current tip-of-tree is? We're still running an old version of this drop-in with the manual changes we made before you merged anything, and it's working great, but I'd love to get rid of the "The Redis object cache drop-in is outdated" message WordPress pesters our support team with by updating the plugin, and not to a specific commit in this repo.

tillkruss:
I'll merge it next week with #97 👍🏻

jack9603301:
Trouble, in cluster mode, this ping command execution detection will lead to redis cluster return errors, but now it seems that the fix bug caused new problems. Is there any other way?

jack9603301:
Under redis-cluster, some commands will not be available.

PR #78, opened 2018-09-17
Added Missing DocBlocks

bhubbard:


tillkruss:
Why do you want this?

bhubbard:
Well I just recently started looking at the code for this plugin. Docblocks are good to encourage documentation and meet WordPress standards. I used a simple shortcode tool to create them, but at least they are in place so they can be better filled out in the future. 

Personally I use [WP Enforcer](https://github.com/stevegrunwell/wp-enforcer) for all of my WordPress projects which uses the WordPress Coding Standards.

PR #87, opened 2018-11-20
igbinary serializer support is added.

skoro:
igbinary serializer is faster than php's serialize() and it also
reduces stored data converting them into a binary form.
This commit brings a new WP_REDIS_IGBINARY constant which enables
igbinary serializer for Redis object cache.


tillkruss:
Nice, thanks!

PR #92, opened 2019-01-08
Adds filter to customize the cache expiration

davefx:
This plugin adds a filter to allowing the customization of the cache expiration setting depending on the inserted key/value in the database.

Using this, we could solve problems like the one described in https://wordpress.org/support/topic/expiration-date-5/

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -600,7 +600,7 @@ class WP_Object_Cache
                 return false;
             }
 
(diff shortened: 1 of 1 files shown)


tillkruss:
I'm not sure about passing the value to the filter. Is there a reason for that?

davefx:
Mmm... Actually I'm not particularly interested in passing the value to the filter, either. I'm going to remove it from the filter call.

PR #93, opened 2019-01-09
Adding redis_cache_expiration filter to set function

davefx:
Yesterday's patch only added the filter to add_or_replace() function. It must be also added to set() function.


tillkruss:
Thanks!

PR #94, opened 2019-01-18
The use of PING command was removed and bug was fixed.

jack9603301:
Redis-Object-Cache plug-in does not support redis-cluster adequately. It is wrong to operate the cluster in the wrong way. You use redis-cluster in the same way as singleton, because the characteristics of redis-cluster determine that some operations are invalid, such as ping, modifying the following code of object-cache.php or diagnostics.php to recover from failure.

- The use of PING command was removed and bug was fixed
- In addition, since redis-cluster mode cannot select database db numbers greater than 0, it is not meaningful to explicitly specify database db number operations in redis-cluster mode. I have changed the relevant parts of redis object cache documents, please update them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,12 +131,6 @@ define( 'WP_REDIS_CLUSTER', [
     'tcp://127.0.0.1:6379?alias=node-01',
     'tcp://127.0.0.2:6379?alias=node-02',
     'tcp://127.0.0.3:6379?alias=node-03',
-    'tcp://127.0.0.4:6379?alias=node-04',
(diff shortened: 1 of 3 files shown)


tillkruss:
This is inadequate. Disabling the ping for cluster mode would preserve the functionality for everyone else.

tillkruss:
Please try these changes I've made.

jack9603301:
Your logical expression is incorrectly written and has been corrected. After online testing, everything is normal.

PR #98, opened 2019-04-26
Handle Redis connection failure gracefully.

### LINKED ISSUE #97: Extend $fail_gracefully to other methods rather than only constructor
**Context:**

The `WP_Object_Cache` class takes `$fail_gracefully` variable as parameter in its constructor. And the Redis Object Cache fails gracefully without throwing any exception if it fails to connect to Redis.

But this does not extend to other methods i.e., `add`, `replace`, `add_or_replace`, `delete`, `get`, `set` etc.

This is fine when the Redis connectivity is checked in the constructor. But let's say if Redis connectivity is lost for some unknown reasons (network failures, Redis server failures etc.) at later stage after constructor is already executed, the plugin will throw an exception and the php process will terminate abruptly.

**Proposal:**
Store the `$fail_gracefully` variable as private class variable. And respect its value in other methods as well along with the constructor such as `add`, `replace`, `set`, `get` etc.

With this, all the methods of `WP_Object_Cache` class will fail gracefully in case of an error.

Open for other suggestions. This scenario was exposed in the production environment of one of the project that I'm working on.

desaiuditd:
Fixes #97

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -295,6 +295,13 @@ class WP_Object_Cache
     private $redis_connected = false;
 
     /**
+     * Check to fail gracefully or throw an exception.
+     *
+     * @var bool
+     */
+    private $fail_gracefully = true;
+
+    /**
      * Holds the non-Redis objects.
      *
      * @var array
@@ -376,6 +383,8 @@ class WP_Object_Cache
     public function __construct($fail_gracefully = true)
     {
         global $blog_id, $table_prefix;
+
+        $this->fail_gracefully = $fail_gracefully;
 
         $parameters = array(
             'scheme' => 'tcp',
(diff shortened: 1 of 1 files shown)


includes/object-cache.php
```diff
@@ -1265,6 +1321,22 @@ protected function is_serialized($data, $strict = true)
...
         return false;
     }
+
+    /**
+     * Handle the redis failure gracefully or throw an exception.
+     *
+     * @param \Exception $exception Exception thrown.
+     */
+    private function handle_redis_failure($exception) {
+        // When Redis is unavailable, fall back to the internal back by forcing all groups to be "no redis" groups
```
aprea: I know this was an existing comment that was simply moved but is `internal back` a typo?

desaiuditd: Fixed - 141b0f7

tillkruss:
I think we should log the silent exception.

tillkruss:
@desaiuditd: What do you reckon about returning early? https://github.com/tillkruss/redis-cache/commit/66431a5c4dcfcd5b28c0fb36860628e5f9a0cd43

desaiuditd:
> @desaiuditd: What do you reckon about returning early? [66431a5](https://github.com/tillkruss/redis-cache/commit/66431a5c4dcfcd5b28c0fb36860628e5f9a0cd43)

Yeah. Looks legit. Explicit early return.

PR #105, opened 2019-06-15
added unflushable groups functionality

naxvog:
- Introduced the ability to add groups that will be ignored once a `wp_cache_flush` is requested
- Added new constant `WP_REDIS_UNFLUSHABLE_GROUPS`
- Changed the lua-script from using the `KEYS` command in favour of the preferable an non blocking `SCAN`

Tested on WordPress 5.2, PHP 7.2 and Redis 5.0.5

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -791,11 +791,19 @@ class WP_Object_Cache
         $salt        = strtr($salt, $rep_squotes);
         if ( ! $this->unflushable_groups ) {
             return <<<EOT
+                local cur = 0
                 local i = 0
-                for _,k in ipairs(redis.call('keys', '{$salt}*')) do
-                    redis.call('del', k)
-                    i = i + 1
-                end
+                local tmp
+                repeat
+                    tmp = redis.call('SCAN', cur, 'MATCH', '{$salt}*')
+                    cur = tonumber(tmp[1])
+                    if tmp[2] then
+                        for _, v in pairs(tmp[2]) do
+                            redis.call('del', v)
+                            i = i + 1
+                        end
+                    end
+                until 0 == cur
                 return i
 EOT;
         }
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -789,6 +789,16 @@ class WP_Object_Cache
     {
         $rep_squotes = array('\'' => '\\\'');
         $salt        = strtr($salt, $rep_squotes);
+        if ( ! $this->unflushable_groups ) {
+            return <<<EOT
+                local i = 0
+                for _,k in ipairs(redis.call('keys', '{$salt}*')) do
+                    redis.call('del', k)
+                    i = i + 1
+                end
+                return i
+EOT;
+        }
         $salt_length = strlen($salt);
         $unflushable = array_map(
             function($group) use ($rep_squotes) {
@@ -797,7 +807,7 @@ class WP_Object_Cache
             $this->unflushable_groups
         );
         $unflushable = '{\'' . implode('\',\'', $unflushable) . '\'}';
-        return "            
+        return <<<EOT          
             local cur = 0
             local i = 0
             local unf = {$unflushable}
@@ -821,7 +831,7 @@ class WP_Object_Cache
                 end
             until 0 == cur
             return i
-        ";
+EOT;
     }
 
     /**
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -807,7 +807,7 @@ EOT;
             $this->unflushable_groups
         );
         $unflushable = '{\'' . implode('\',\'', $unflushable) . '\'}';
-        return <<<EOT          
+        return <<<EOT
             local cur = 0
             local i = 0
             local unf = {$unflushable}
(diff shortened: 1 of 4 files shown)


tillkruss:
Why is `userlogins` not flushed by default?

naxvog:
Just had it in there from the [issue referring to group based flushing](https://github.com/tillkruss/redis-cache/issues/76) and for testing purposes. Removed it as it is a remnant.

includes/object-cache.php
```diff
@@ -699,14 +710,7 @@ public function flush($delay = 0)
...
             if ($salt && $selective) {
-                $script = "
-                    local i = 0
-                    for _,k in ipairs(redis.call('keys', '{$salt}*')) do
-                        redis.call('del', k)
-                        i = i + 1
-                    end
-                    return i
-                ";
+                $script = $this->lua_flush_script($salt);
```
tillkruss: Can we split this into two methods, one for selective, non-blocking flush and one for the default block flush when `$unflushable_groups` is empty?

includes/object-cache.php
```diff
@@ -775,6 +779,51 @@ public function flush($delay = 0)
...
+        $salt        = strtr($salt, $rep_squotes);
+        $salt_length = strlen($salt);
+        $unflushable = array_map(
+            function($group) use ($rep_squotes) {
+                return ':' . strtr($group, $rep_squotes) . ':';
+            },
+            $this->unflushable_groups
+        );
+        $unflushable = '{\'' . implode('\',\'', $unflushable) . '\'}';
+        return "
```
tillkruss: Can we use a _heredoc_ for the Lua code?

includes/object-cache.php
naxvog: I would incorporate this functionality in the function itself.
Regarding the block flush: [using `KEYS` in production environments is discouraged](https://redis.io/commands/KEYS)

How about this flush as substitute if `$unflushable_groups` is empty?
```php
local cur = 0
local i = 0
local tmp
repeat
    tmp = redis.call('SCAN', cur, 'MATCH', '{$salt}*')
    cur = tonumber(tmp[1])
    if tmp[2] then
        for _, v in pairs(tmp[2]) do
            redis.call('del', v)
            i = i + 1
        end
    end
until 0 == cur
return i
```

tillkruss: As long as it's blocking, whatever works best. I'd like to avoid possible race conditions with the blocking call.

naxvog: Redis guarantees that an `EVAL` call is atomic so according to the docs [all lua scripts are blocking](https://redis.io/commands/eval#atomicity-of-scripts).

So why use `SCAN` in the first place? According to [this benchmark in the redis repo](https://github.com/antirez/redis/issues/3190#issuecomment-215014512) `SCAN` is faster in lua scripts by a significant factor. Although I could not verify the results on my redis 5 installation it might be an issue for older versions.

Theoretically flushing with `SCAN` will decrease the total amount of database-entries with every page returned. Finding the next page will therefore be generally faster than the last page.

tillkruss: I'm easy. If it's faster and backwards compatible, then let's switch 👌🏻

tillkruss:
@naxvog: Is this ready to be merged?

naxvog:
I do think so.
Keep in mind that the `SCAN` command is available since redis 2.8.0

tillkruss:
@naxvog: I've removed the quote stuff, LMK if `master` works for you:

https://github.com/tillkruss/redis-cache/commit/57c8ed6edfe9efcb464a9fca575f659fa94c8b89

naxvog:
Generally it might be a problem to use the salt and groups unescaped as it can contain LUA so an injection would be possible as far as I know. The best solution would be to use arguments on the LUA scripts to let redis handle the escaping.

The other issue is not as severe but it might be a problem non the less: it would be possible to craft a salt that allows the flashing of entries that should not be flashed using a salt like `abc:[group-name]`. This would also be a problem with certain groups that contain `:`.
A fix would be to replace `:` with another character in `WP_Object_Cache::build_key` as well as escaping the salt in the new `lua_flush_script` method or before in `WP_Object_Cache::flush`

tillkruss:
> Generally it might be a problem to use the salt and groups unescaped as it can contain LUA so an injection would be possible as far as I know.

I'm not concerned about injections. If someone can define PHP constants, then it's game over anyhow.

> The best solution would be to use arguments on the LUA scripts to let redis handle the escaping.

Can you submit a PR for that? The backslash quotes stuff will be hard to understand in 6 months from now.

> The other issue is not as severe but it might be a problem non the less: it would be possible to craft a salt that allows the flashing of entries that should not be flashed using a salt like `abc:[group-name]`. > This would also be a problem with certain groups that contain `:`.
> A fix would be to replace `:` with another character in `WP_Object_Cache::build_key` as well as escaping the salt in the new `lua_flush_script` method or before in `WP_Object_Cache::flush`

Can you submit a fix for that as well?

naxvog:
LUA arguments: https://github.com/tillkruss/redis-cache/pull/109

PR #110, opened 2019-07-06
Null key separator

naxvog:
Problem:
Using the `:` as separator building the redis key might cause issues if a group name contains that separator.

Solution:
Researched a bit and came to the conclusion that it would be best to have the keys separated by the null-char as its usage should be far less than `:` and you might argue that removing a null-char during key-sanitization is a good thing.

Redis is handling that key just fine using `\x00` so retrieving the value is not an issue.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1144,7 +1144,10 @@ LUA;
         $salt = defined('WP_CACHE_KEY_SALT') ? trim(WP_CACHE_KEY_SALT) : '';
         $prefix = in_array($group, $this->global_groups) ? $this->global_prefix : $this->blog_prefix;
 
-        return "{$salt}{$prefix}:{$group}:{$key}";
+	    $sep = chr(0);
+	    $group = strtr( $group, $sep, '' );
+
+        return "{$salt}{$prefix}{$sep}{$group}{$sep}{$key}";
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tillkruss:
What does WordPress use by default for group separators?

naxvog:
WordPress uses a associative array based non-persistent storage:
`$this->cache[ $group ][ $key ] = $data;`

Unfortunately as far as I know it is impossible to use redis data structures as only strings support automatic expiry

tillkruss:
Mhh. I much rather translate `:` in group names to something else (maybe `_`), than use null characters, just to keep the Redis keys easy to read for humans.

> Unfortunately as far as I know it is impossible to use redis data structures as only strings support automatic expiry

I know that pain.

naxvog:
Well translating the separating character might introduce collisions with other group names.
Readability isn't too bad to be honest as redis is translating `char(0)` to `\x00` (live example on one if my sites; lines truncated; yes my salt is overkill):
```
1562686562.060375 [1 unix:/var/run/redis/redis.sock] "SETEX" "1K1~tROJ8AjFiggNG93kLqFHDeeE1xqNFIDO_VC2R7L0fa_:options:notoptions" "2592000" "a:61:{s:2...
1562686562.060375 [1 unix:/var/run/redis/redis.sock] "SETEX" "1K1~tROJ8AjFiggNG93kLqFHDeeE1xqNFIDO_VC2R7L0fa_\x00options\x00notoptions" "2592000" "a:61:{s:2...
```

tillkruss:
I love the null char solution, but let's stick with a simple replacement until we actually encounter a conflict.

https://github.com/tillkruss/redis-cache/commit/71fe0efad2213947be66af3b821766ece74ed68b

PR #111, opened 2019-07-16
Add config options to specify timeout

### LINKED ISSUE #107: Need a way to specify timeout and retry_interval
I am using redis-cache in a quite big production installation, paired with the pecl client (phpredis).

The latter allows specifying a few parameters like `timeout` and `retry_interval` (ref https://github.com/phpredis/phpredis#connect-open), and I'd very much like to do so, as the defaults are **0** which means **hang forever**.

I am not sure whether other non-pecl clients would allow these parameters, but again it would be really useful to have them in place.

I can help with pull-requests & friends

kjlaw89:
This change adds 3 new constants: `WP_REDIS_TIMEOUT`, `WP_REDIS_READ_TIMEOUT`, `WP_REDIS_RETRY_INTERVAL`. It also defaults the timeouts to 5 seconds, which may be a bit high but is better than infinite!

This should hopefully resolve #107. We actually just ran into this issue ourselves the other day. Unfortunately we've only got PECL-Redis installed - I do not have access to a Predis or HHVM Redis configuration to test.


tillkruss:
Rad! You can read about the Predis parameters here: https://github.com/nrk/predis/wiki/Connection-Parameters

After that's added I can merge it right away.

kjlaw89:
Lines 518 and 519 should cover it for Predis (`timeout` is already provided, `read_timeout` in their config is `read_write_timeout`). I didn't think I had an environment setup for testing it, but I just realized it's included with the plugin. It tested it in our dev environment and it worked as expected.

tillkruss:
Ah, got it.

Can you set Predis as the client and make sure it works? It's bundled with the plugin.

kjlaw89:
Yep, worked as expected.

PR #118, opened 2019-09-30
Fix multi get

villesiltala:
* Fix an error in `get_multi()`. If no values are found for given keys, the method sets empty values to prevent errors in array handling.


villesiltala:
Again, this should be a PR to our fork. I never remember GitHub defaults PRs in forks to the original repository. 😅

PR #124, opened 2019-11-02
Plesk phpredis fix

naxvog:
As mentioned in https://github.com/tillkruss/redis-cache/issues/123 plesk ships with phpredis version 3.0.0. As this version only supports 5 arguments for the redis `connection` method trying to call it with the 6th argument (retry_timeout) will result in a silent connection failure.

Digging into the phpredis source I found that the 6th argument was introduced in version 3.1.3 so this fix checks the installed version to that specific one and adds the 6th argument only if applicable.


tillkruss:
Thanks!

Issue #126, opened 2019-11-13
Flush don't work with WP_REDIS_SELECTIVE_FLUSH set to TRUE


kionez:
With WP_REDIS_SELECTIVE_FLUSH set to TRUE, flushing cache don't work as expected.
I have to use WP_REDIS_SELECTIVE_FLUSH because I have many different wordpress installation uplodaed by users and I must definitely keep them "separate".

After some digging in the code and a bit of redis-cli (expecially MONITOR command), i see that the LUA script was not executed, if I try to copy-paste and run manually it gives me an error:

```
127.0.0.1:6379> "EVAL" " local cur = 0\n local i = 0\n local tmp\n repeat\n tmp = redis.call('SCAN', cur, 'MATCH', 'MYDOMAIN.com*')\n cur = tonumber(tmp[1])\n  if tmp[2] then\n for _, v in pairs(tmp[2]) do\n redis.call('del', v)\n i = i + 1\n  end\n end\n until 0 == cur\n return i" "0"

(error) ERR Error running script (call to f_523831630e0cc0dbf20c9141913c81f1bec591ba): @user_script:9: @user_script: 9: Write commands not allowed after non deterministic commands. Call redis.replicate_commands() at the start of your script in order to switch to single commands replication mode.
```

So, i edit my wp-content/object-cache.php adding `redis.replicate_commands()` at the start of LUA script (lines  820 and 860) and now cache is correctly flushed (via dashboard button or via `wp cache flush` command)

I don't know if this is the correct solution, I don't have much knowledge of redis, I try to read https://redis.io/commands/eval but I can't figure out.

Hope it helps!

Here some info:
Wordpress: 5.3 and 5.2
Redis Object Cache: 1.5.0
Client: PhpRedis (v4.3.0)
Redis Extension: 4.3.0
Predis Client: 1.1.1
PHP Version: 5.6.40-0+deb8u6
Redis Object Cache Drop-In v1.5.0 by Till Krüss
Redis Server: 3.2.8

tillkruss:
Thanks for the report! Can you submit a PR with the fix?

kionez:
I've search for more info about `redis.replicate_commands()` and it seems a correct way to fix this kind of situation. Hope it helps :)

PR #127, opened 2019-11-14
Flush don't work with WP_REDIS_SELECTIVE_FLUSH set to TRUE (issue #126)

kionez:
This PR fix the LUA error, reported in issue #126 

`(error) ERR Error running script (call to f_523831630e0cc0dbf20c9141913c81f1bec591ba): @user_script:9: @user_script: 9: Write commands not allowed after non deterministic commands. Call redis.replicate_commands() at the start of your script in order to switch to single commands replication mode.`

PS: I don't know if this is the correct solution, I don't have much knowledge of redis, I try to read https://redis.io/commands/eval but I can't figure out.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -845,6 +845,7 @@ class WP_Object_Cache
     {
         return function () use ($salt) {
             $script = <<<LUA
+                redis.replicate_commands()
                 local cur = 0
                 local i = 0
                 local tmp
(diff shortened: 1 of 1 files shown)


naxvog:
Keep in mind that `redis.replicate_commands()` should only be needed in redis versions between 3.2 and 5.0. The documentation states that the command is available in 3.2 and the default behavior since 5.0 ([documentation](https://redis.io/commands/eval#replicating-commands-instead-of-scripts))

I am fairly certain that the `redis.replicate_commands()` command call in versions before 3.2 would cause issues.

Will try to add a PR asap

PR #128, opened 2019-11-15
added lua script redis version check

naxvog:
As mentioned in https://github.com/tillkruss/redis-cache/pull/127

Added a version check to the `redis.replicate_commands()` LUA command as it is not needed in redis versions prior to 3.2 and since 5.0.
Additionally the actual redis version was not tracked before so I added the functionality and displayed it in the diagnostics section.


tillkruss:
Excellent work my friend!

PR #134, opened 2019-11-28
Removed strtolower in build key method to fix LUA scripts

naxvog:
Fix for https://github.com/tillkruss/redis-cache/issues/130

Chose the simpler method of removing the `strtolower` function as it will lower complexity in favor of transforming the salt in the LUA scripts as well.


tillkruss:
This will cause a massive spike in `MISSES` and leave unused keys for installations without eviction policy or maxTTL.

**Everyone should flush their cache after updating to v1.5.5.**

PR #141, opened 2019-12-10
Added object cloning for the in-memory cache

naxvog:
Added object cloning before local in-memory cache insertion as per https://github.com/tillkruss/redis-cache/issues/140

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1316,6 +1316,9 @@ LUA;
      */
     public function add_to_internal_cache($derived_key, $value)
     {
+        if (is_object($value)) {
+            $value = clone $value;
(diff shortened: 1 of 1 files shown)


tillkruss:
Thanks!

We’d need the same logic when retrieving internal objects to avoid the cache being updated by reference.

naxvog:
Objects are already cloned upon retrieval as per https://github.com/tillkruss/redis-cache/blob/1c133aaf3bb17339afb727a5c43fe25629f9dfcf/includes/object-cache.php#L973

tillkruss:
Fantastic. I’m on a business trip this week. Will tag new version early next week.

PR #142, opened 2019-12-16
Ping on cluster using Predis

jack9603301:
Redis Cluster does not support the Ping command

### DIFF
--- a/includes/diagnostics.php
+++ b/includes/diagnostics.php
@@ -10,7 +10,7 @@ $info[ 'Status' ] = $this->get_status();
 if (defined( 'WP_REDIS_CLIENT' ) && WP_REDIS_CLIENT === 'predis' && defined( 'WP_REDIS_CLUSTER' )) {
     $info[ 'Redis Version' ] = 'Not supported';
 } else {
-    $info[ 'Redis Version' ] = $this->get_redis_version() ?: 'Unknown';
+    $info[ 'Redis Version' ] = $this->get_redis_version();
 }
 $info[ 'Client' ] = $this->get_redis_client_name();
 
--- a/includes/diagnostics.php
+++ b/includes/diagnostics.php
@@ -3,27 +3,20 @@
 global $wp_object_cache;
 
 $info = $plugins = $dropins = array();
-$dropin = $this->validate_object_cache_dropin() && ( ! defined('WP_REDIS_DISABLED') || ! WP_REDIS_DISABLED );
+$usesPredis = defined( 'WP_REDIS_CLIENT' ) && WP_REDIS_CLIENT === 'predis';
+$isCluster = defined( 'WP_REDIS_CLUSTER' );
+$isDisabled = defined( 'WP_REDIS_DISABLED' ) && WP_REDIS_DISABLED;
+
+$dropin = $this->validate_object_cache_dropin() && ! $isDisabled;
 
 $info[ 'Status' ] = $this->get_status();
-
-$info[ 'Redis Version' ] = (defined( 'WP_REDIS_CLIENT' ) && WP_REDIS_CLIENT === 'predis' && defined( 'WP_REDIS_CLUSTER' ))?'Not supported':$this->get_redis_version();
-/*
-if (defined( 'WP_REDIS_CLIENT' ) && WP_REDIS_CLIENT === 'predis' && defined( 'WP_REDIS_CLUSTER' )) {
-    $info[ 'Redis Version' ] = 'Not supported';
-} else {
-    $info[ 'Redis Version' ] = $this->get_redis_version();
-}
-*/
+$info[ 'Redis Version' ] = $this->get_redis_version() ?: 'Unknown';
 $info[ 'Client' ] = $this->get_redis_client_name();
-
 $info[ 'Drop-in' ] = $dropin ? 'Valid' : 'Invalid';
-
-$usesPredis = defined( 'WP_REDIS_CLIENT' ) && WP_REDIS_CLIENT === 'predis';
 
 if ( $dropin ) {
     try {
-        if ($usesPredis && defined( 'WP_REDIS_CLUSTER' )) {
+        if ($usesPredis && $isCluster) {
             $info[ 'Ping' ] = 'Not supported';
         } else {
             $cache = new WP_Object_Cache( false );
--- a/includes/diagnostics.php
+++ b/includes/diagnostics.php
@@ -6,14 +6,19 @@ $info = $plugins = $dropins = array();
 $dropin = $this->validate_object_cache_dropin() && ( ! defined('WP_REDIS_DISABLED') || ! WP_REDIS_DISABLED );
 
 $info[ 'Status' ] = $this->get_status();
-$info[ 'Redis Version' ] = $this->get_redis_version() ?: 'Unknown';
+
+if (defined( 'WP_REDIS_CLIENT' ) && WP_REDIS_CLIENT === 'predis' && defined( 'WP_REDIS_CLUSTER' )) {
+    $info[ 'Redis Version' ] = 'Not supported';
+} else {
+    $info[ 'Redis Version' ] = $this->get_redis_version() ?: 'Unknown';
+}
 $info[ 'Client' ] = $this->get_redis_client_name();
 
 $info[ 'Drop-in' ] = $dropin ? 'Valid' : 'Invalid';
 
 if ( $dropin ) {
     try {
-        if ( defined( 'WP_REDIS_CLUSTER' ) ) {
+        if (defined( 'WP_REDIS_CLIENT' ) && WP_REDIS_CLIENT === 'predis' && defined( 'WP_REDIS_CLUSTER' )) {
             $info[ 'Ping' ] = 'Not supported';
         } else {
             $cache = new WP_Object_Cache( false );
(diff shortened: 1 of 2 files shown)


tillkruss:
I haven't gotten any reports on this. Why was it merged in #70? 

What client and version are you using?

jack9603301:
I use the latest redis cluster and predis library to link. I don't know why I was contracted in #70, but in my environment, the ping command can't be submitted to the cluster by predis!

tillkruss:
I won’t be supporting Predis much longer. Can you please test this using PhpRedis.

jack9603301:
The following error occurred after I changed to PECL:
```
Couldn't map cluster keyspace using any provided seed (RedisClusterException)
```

tillkruss:
It looks like you're trying to use cluster mode, but your running a regular Redis instace.

Does using `redis-cli -c -p 6380` work as expected?

jack9603301:
My redis cluster uses the following command to access normally:

```
redis-cli -a password -c -p port 
```

tillkruss:
Ping is working with PhpRedis, feel free to update this PR only disabling ping on clusters when Predis is used.

jack9603301:
But I already did what you said in this PR, please let me know if I make any mistake on this PR, otherwise, you can merge this PR.

jack9603301:
please let me know if I make any mistake on this PR, otherwise, you can merge this PR.

includes/object-cache.php
```diff
@@ -566,13 +566,14 @@ public function __construct($fail_gracefully = true)
                 $this->redis_client .= sprintf(' (v%s)', Predis\Client::VERSION);
             }
 
-            if (defined('WP_REDIS_CLUSTER')) {
-                $this->redis->ping(current(array_values(WP_REDIS_CLUSTER)));
+            if ((strcasecmp('predis', $client) === 0) && defined('WP_REDIS_CLUSTER')) {
+                // Redis Cluster does not support the Ping command
+                //$this->redis->ping(current(array_values(WP_REDIS_CLUSTER)));
```
tillkruss: This will break pings for non Predis clusters.

includes/object-cache.php
```diff
@@ -566,13 +566,14 @@ public function __construct($fail_gracefully = true)
...
             } else {
                 $this->redis->ping();
             }
 
-            if ( ! isset( $options['replication'] ) || ! $options['replication'] ) {
+            if ( ((strcasecmp('predis', $client) === 0) && !defined('WP_REDIS_CLUSTER')) && (! isset( $options['replication'] ) || ! $options['replication']) ) {
```
tillkruss: These should be extracted into variables, such as `$usesReplication` and `$onPredisCluster`.

includes/diagnostics.php
```diff
@@ -6,15 +6,24 @@
...
+}
 $info[ 'Client' ] = $this->get_redis_client_name();
 
 $info[ 'Drop-in' ] = $dropin ? 'Valid' : 'Invalid';
 
 if ( $dropin ) {
     try {
-        $cache = new WP_Object_Cache( false );
-        $info[ 'Ping' ] = defined( 'WP_REDIS_CLUSTER' ) ? 'Not supported' : $cache->redis_instance()->ping();
+        if (defined( 'WP_REDIS_CLIENT' ) && WP_REDIS_CLIENT === 'predis' && defined( 'WP_REDIS_CLUSTER' )) {
```
tillkruss: This is redundant. Should be extracted into `$usesPredis` boolean.

includes/diagnostics.php
```diff
@@ -6,15 +6,24 @@
 $dropin = $this->validate_object_cache_dropin() && ( ! defined('WP_REDIS_DISABLED') || ! WP_REDIS_DISABLED );
 
 $info[ 'Status' ] = $this->get_status();
-$info[ 'Redis Version' ] = $this->get_redis_version() ?: 'Unknown';
+
+if (defined( 'WP_REDIS_CLIENT' ) && WP_REDIS_CLIENT === 'predis' && defined( 'WP_REDIS_CLUSTER' )) {
+    $info[ 'Redis Version' ] = 'Not supported';
+} else {
+    $info[ 'Redis Version' ] = $this->get_redis_version() ?: 'Unknown';
```
tillkruss: Wouldn't this just default to `Unknown`? Is this condition really necessary.

jack9603301: Original judgement
```
$info[ 'Redis Version' ] = $this->get_redis_version() ?: 'Unknown';
```
It may not work, so I made a change, in fact, it does have problems in my environment! And not in line with the general way of judgment! The format of ternary logic judge is:
```
boolean?true:false
```
So the part of true should not be empty

jack9603301:
Please continue to review and feedback the existing code, otherwise, it can be merged!

includes/diagnostics.php
```diff
@@ -6,15 +6,26 @@
 $dropin = $this->validate_object_cache_dropin() && ( ! defined('WP_REDIS_DISABLED') || ! WP_REDIS_DISABLED );
 
 $info[ 'Status' ] = $this->get_status();
-$info[ 'Redis Version' ] = $this->get_redis_version() ?: 'Unknown';
+
+if (defined( 'WP_REDIS_CLIENT' ) && WP_REDIS_CLIENT === 'predis' && defined( 'WP_REDIS_CLUSTER' )) {
+    $info[ 'Redis Version' ] = 'Not supported';
+} else {
+    $info[ 'Redis Version' ] = $this->get_redis_version();
+}
```
tillkruss: Why is this necessary? What is the exact error you're seeing without it?

jack9603301: 1. The preferred condition to check whether version is available should not be determined by whether the normal value of `$this->get_redis_version()` is true or false
2. Boolean expression in the form: 
```
Boolean? True: false
```

as well as

```
$this->get_redis_version()?:"Unkwon";
```

true or false should be returned for `$this->get_redis_version()`, not other return types, and should avoid the possibility of throwing exceptions!

3. For a Boolean expression, the execution logic is that when Boolean returns a Boolean value, the code of true or false segment and its return value are executed and returned according to the Boolean value. Therefore, the true segment should not be empty!

4. Maybe my writing method is not the only one, but it can guarantee that there will be no serious problems in logic! In fact, a conforming Boolean expression should have the same effect on its equivalent code. Any conforming Boolean expression can be converted to the form of `if-else-endif` normally。

When I modified the bug, I found that redis cluster was unable to execute the server command to obtain version information. After rewriting the judgment during debugging and running normally, I decided to modify the judgment logic of the location based on the above reasons!

tillkruss: It's safe cast strings as booleans: https://www.php.net/manual/en/types.comparisons.php

What's the exception you are seeing without your fix?

jack9603301: It is not a matter of type conversion, but of the meaning of Boolean expressions, such as:
```
$tmp = fun() ? 1:2
```
The fun function will be called. If the return value of the function is true, the expression will return 1, otherwise it will return 2. In the original written expression, there is no true condition specified for the Boolean expression. Moreover, if the Boolean condition of the expression is function function call, once the function throws an exception, it will not execute normally!

It's been several days since the code was modified. If I want to reproduce it, I have to modify it again. It's troublesome, so I won't reproduce it. You can try it yourself! Since the later modified code can run normally and meet the specification, it is recommended to keep it!

tillkruss: Why would `get_redis_version()` ever throw an exception?

jack9603301:
In fact, I don't think the key is to throw an exception, because the original writing method is:

```
$info[ 'Redis Version' ] = $this->get_redis_version() ?: 'Unknown';
```

If `$this - > get_redis_version() ` returns true, the expression will have no return value!

tillkruss:
> If `$this - > get_redis_version() ` returns true, the expression will have no return value!

Exactly, unless a truthy value is returned, that expression will show `Unknown`, that's you can undo that change 👍🏻

jack9603301:
I have modified the logic and reduced it to the equivalent writing method of logical expression. Can you see?

includes/object-cache.php
```diff
@@ -566,13 +566,20 @@ public function __construct($fail_gracefully = true)
...
+                }
             } else {
                 $this->redis->ping();
             }
 
-            if ( ! isset( $options['replication'] ) || ! $options['replication'] ) {
+            $usesReplication = isset($options['replication']) && $options['replication'];
+
+            if (! $usesReplication) {
                 $server_info = $this->redis->info( 'SERVER' );
```
tillkruss: @jack9603301: Can you please test this code with your Predis cluster and post the exception it throws, if it throws any?

jack9603301: There is no judgment. I have submitted relevant corrections. Please see what I submitted. The reason why I made a judgment is that I found that the `server` instruction may have execution errors in predis mode. It is forbidden to execute the command for redis cluster.

tillkruss: I will not merge this pull request, unless I understand it and see the corresponding exception/error message.

jack9603301: I remember that I gave you an exception about ping command execution error, which is the same as the exception!

jack9603301: I remember that I gave you an exception about ping command execution error, which is the same as the exception!

tillkruss:
I don't feel comfortable merging this pull request until exception messages for all code changes are provided.

PR #145, opened 2020-01-15
hide notice on missing read_timeout parameter

edubacco:
If notices are shown on pages, the application can break due to "headers already sent" PHP error.
This fix the problem.


tillkruss:
Thanks!

edubacco:
You're welcome, your plugin is great for our infrastructure =)

Issue #146, opened 2020-02-04
Add PHPRedis (pecl) TLS Support


timofeycom:
Hi Till, per our conversation on [WP Support](https://wordpress.org/support/topic/bug-phpredis-pecl-does-not-work-with-tls-scheme/) please add TLS support for PHPRedis (pecl). I've created a patch for this.

# Patch
Rename to patch.diff
[patch.diff.txt](https://github.com/tillkruss/redis-cache/files/4151094/patch.diff.txt)

# Description of the Issue
Redis-Cache 1.5.6 fails to check for `WP_REDIS_SCHEME` when connecting via PHPRedis.

PHPRedis allows TLS connection in the form of tls://127.0.0.1, see https://github.com/phpredis/phpredis#example-1

# Solution / Description of the Patch
This patch add a line of code that checks for `WP_REDIS_SCHEME` when connecting via PHPRedis.

I am currently using this patch to successfully connect to DigitalOcean Redis managed database via TLS.

Let me know if you need anything to get this implemented.

Thank you!
Tim

tillkruss:
@timofeycom: Can you submit a PR here on GitHub, instead of the patch file?

timofeycom:
@tillkruss I'm getting a permissions error 403 when I try to submit Pull Request to the master. It only allows me to submit it to the page-cache branch. Am I missing something?

Thanks

tillkruss:
I see your fork, you'd need to commit to a branch on https://github.com/timofeycom/redis-cache.

PR #147, opened 2020-02-04
Add TLS support

tillkruss:
Based on @timofeycom's patch from #146.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -478,6 +478,10 @@ class WP_Object_Cache
                         $parameters['retry_interval'],
                     ];
 
+                    if ($parameters['scheme'] == 'tls') {
+                        $connection_args[0] = $parameters['scheme'] . '://' . $parameters['host'];
+                    }
+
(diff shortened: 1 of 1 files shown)


tillkruss:
@timofeycom: We need to document the new option. Can you add documentation for the new option?

timofeycom:
@tillkruss as far as the `readme.txt` file, that shouldn't change. It's still stands current.

Not sure how you perfer to document on the technical side. I can leave the notes here.

The new option considers the `WP_REDIS_SCHEME` constant before connecting via PHPRedis. This allows users to connect to a Redis database using a secure connection.

PHPRedis added support for tls connections in [version 5.0](https://github.com/phpredis/phpredis#connect-open). It can now be connected via `$redis->connect('tls://127.0.0.1');`.

This option adds the `tls` protocol to PHPRedis connections only if `WP_REDIS_SCHEME` value has been set to `tls`.

To connect to via PHPRedis using a secure connection, users must have PHPRedis 5.0 or above.

PR #156, opened 2020-05-11
Optional HTML footer with stats/diagnostics

### LINKED ISSUE #154: Set HTTP Response Header to show cached response
Hello Till,

i really appreciate your work on this plugin and i already use it successfully on my development page.
However i have a suggestion: When uing the plugin, you can clearly see which reponse was served by redis, as the speed is by far higher. However, a HTTP-Header indicating a cached response would be great.
Do you have an idea how to implement this? Do you think this could be included in this plugin?
Suggestion:
X-Redis-Cache: HIT
X-Redis-Cache: MISS
X-Redis-Cache: BYPASS (If that is possible/necessary)

What do you think?

Best regards!

tillkruss:
- [x] Show stats when `WP_DEBUG` is enabled
- [ ] Show stats via `WP_REDIS_*` constant
- [ ] Show stats via secure URL parameter
- [ ] Split hits/missing: Redis and in-memory
- [ ] Refine comment template (one liner?)

Resolves #154

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+- Fixed MaxTTL warning condition
+
 ## 1.5.8
 
 - Added warning message about invalid MaxTTL
--- a/includes/admin-page.css
+++ b/includes/admin-page.css
@@ -5,36 +5,6 @@
     padding-bottom: 7px;
 }
 
-.wrap .button.delete {
-    background: #d54e21;
-    border-color: #a83d1a;
-    color: white;
-    -webkit-box-shadow: inset 0 1px 0 #e68260, 0 1px 0 rgba(0, 0, 0, 0.15);
-    box-shadow: inset 0 1px 0 #e68260, 0 1px 0 rgba(0, 0, 0, 0.15);
-}
-
-.wrap .button.delete:hover,
-.wrap .button.delete:focus {
-    background: #c7491f;
-    border-color: #923517;
-    color: white;
-}
-
-.wrap .button.delete:focus {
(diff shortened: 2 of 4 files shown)


lkraav:
I wonder if integration with Query Monitor is in scope here? cc @johnbillion

tillkruss:
@lkraav: We should probably just add support for this: https://github.com/johnbillion/query-monitor/blob/3402629d84ad2ce6ede80f60762b68fe319945cb/output/html/overview.php#L186-L210

@johnbillion: Unless we want to add a dedicated section for the Object Cache with heaps of details, such as cache groups, total size, etc.

johnbillion:
The reason QM doesn't display detailed info about the cache stats is because, as you'll probably know, the WP cache API doesn't provide abstraction for this, so each cache controller stores its stats in a different place.

Let me know if you need me to add anything to QM to enable more detailed stats from the cache controller. I don't think there's a hook in there at the moment to add data to the overview, but you can add your own panel.

tillkruss:
@johnbillion: Yeah, the Pro version adds it's own panel.

How do you feel about about setting a standard for an object cache panel? QM already uses the `stats` and `cache_hit_percentage` anyhow.

johnbillion:
I like that idea. QM could define an object cache panel and cache controllers push their data into it regardless of how they internally store stats.

tillkruss:
Let's move this to https://github.com/johnbillion/query-monitor/issues/518/.

tillkruss:
Closing in favor of #167.

tillkruss:
See https://github.com/tillkruss/redis-cache/issues/154#issuecomment-636100823 for a summary of footer comments on other caching plugins.

PR #165, opened 2020-05-25
Added group sanitization

naxvog:
as per https://github.com/tillkruss/redis-cache/issues/164

Implemented group sanitization so that via configuration constants defined groups do appear sanitized in the plugin diagnostics section.
While implementing I also added some utility methods that should improve code readability: `is_ignored_group`, `is_global_group` and `is_unflushable_group`

Additionally it might be a good idea to implement some kind of warning in the settings page that a group using an illegal character was defined.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -419,15 +419,24 @@ class WP_Object_Cache
         }
 
         if (defined('WP_REDIS_GLOBAL_GROUPS') && is_array(WP_REDIS_GLOBAL_GROUPS)) {
-            $this->global_groups = WP_REDIS_GLOBAL_GROUPS;
+            $this->global_groups = array_map(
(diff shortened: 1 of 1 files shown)


tillkruss:
I'm slammed right now 😓

I'm gonna try to tackle this on the weekend. Thanks for the PR ❤️

tillkruss:
This is great work man, the only downside of this is that we're adding more internal method calls 😐

naxvog:
Correct me if I'm wrong but why are more internal method calls a bad thing? Opcache should cache method calls done with the same parameters if they aren't using any random elements regardless.
Sure there might be the call stack that gets populated but I don't think that this will impact performance at all.

tillkruss:
The time adds up after 1000+ calls. Blackfire.io visualizes and shows the timing nicely.

Issue #166, opened 2020-05-26
Race conditions in function add_or_replace


ffxluca:
My fellas are using the drop-in replacement in a quite big WordPress (single site) setup:
- 10 different apache webservers, each one running on a separate machine/vm.
- php-fpm (`pm` set to `dynamic`, up to 8 children each)

This is almost impossible to replicate but I have noticed the function `add_or_replace()` utilises `exists($derived_key);` and then `set()` or `setex()`, whereas it should instead utilise Redis' [setnx](https://redis.io/commands/setnx).

I am experimenting on a local branch (with php-redis version ~4 or ~5, and pecl client) so I'm going to open a pull request soon - that'll make it easier to understand what I'm trying to avoid

tillkruss:
Are you encountering [this](https://core.trac.wordpress.org/ticket/31245)?

ffxluca:
@tillkruss not really, we already have a workaround for that.

I cannot open new branches here so I cloned the repo, this is more or less the diff
https://github.com/ffxluca/redis-cache/commit/d1a528586e50b761152d98da08c227087c7c9eb3

tillkruss:
Alright, can you submit a PR after you finished your testing?

ffxluca:
Deffo 😃 I might need some help with testing the `predis` client and possibly the permission to push new branches here

naxvog:
Just create a pull request from your own forked repository. You don't need any permission in this repo to do this.
Predis support is being dropped soon so I would not bother to test specifically for it (see https://github.com/tillkruss/redis-cache/issues/160).

tillkruss:
Thanks @naxvog.

I won't drop Predis support until WP drops PHP 7.4, just switch over to a new default client when PhpRedis isn't installed.

tillkruss:
@ffxluca: Were you able to test your fix?

ffxluca:
@tillkruss yes, tested as in _nothing broke here_ I'll open a pull request and make more thorough tests :)

PR #170, opened 2020-05-29
Use setnx in add_or_replace() function when the $add argument is set

ffxluca:
Fix for #166

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -709,17 +709,23 @@ class WP_Object_Cache
         // save if group not excluded and redis is up
         if (! in_array($group, $this->ignored_groups) && $this->redis_status()) {
             try {
-                $exists = $this->redis->exists($derived_key);
-
-                if ($add == $exists) {
-                    return false;
-                }
-
                 $expiration = apply_filters('redis_cache_expiration', $this->validate_expiration($expiration), $key, $group);
 
-                if ($expiration) {
+                if ($add) {
+                    // use `setnx` (atomic operation) and then extend the TTL
+                    $result = $this->parse_redis_response($this->redis->setnx($derived_key, $this->maybe_serialize($value)));
+                    if (! $result ) {
+                        return false;
+                    }
+
+                    if ($expiration) {
+                        $this->redis->expire($derived_key, $expiration);
+                    }
+                } elseif ($expiration) {
+                    // update value and extend TTL
                     $result = $this->parse_redis_response($this->redis->setex($derived_key, $expiration, $this->maybe_serialize($value)));
                 } else {
+                    // only update the value
                     $result = $this->parse_redis_response($this->redis->set($derived_key, $this->maybe_serialize($value)));
                 }
             } catch (Exception $exception) {


tillkruss:
@naxvog: Any thoughts?

ffxluca:
Let me try and put some random `sleep()` statements to actually see the `setnx` returning false

naxvog:
Just to reiterate:
* There is a race condition between checking for existence of a key and setting the key while multiple WordPress instances (clients) trying to add a key at the same time.
* The race condition occurs if client A checked the existence of the key and has not yet set the key while client B does the same existence-check. Both clients will set the key in this case resulting in an error due to the `SETEX` command being called on a value that already exists overriding that value.
* In the example above client A should set the key, client B should receive `false` as the key already exists.

The fix seems fine to me. Nice catch!

However as far as I know both `SETEX` and `SETNX` are atomic redis operations so the comments are a bit misleading. The problem that introduces the race condition is the existence check not the command itself.

On the performance side this should perform better than the old implementation as the existence check is done by `SETNX` and the expiry is just set if it is required. So there is a minimum of 1 and a maximum of 2 redis commands sent whereas before there were always 2 being sent while adding a key.

tillkruss:
Can we use `SET $key $value NX EX $expiration` instead of the extra `EXPIRE` call?

naxvog:
To quote the redis documentation for the options to the `SET` command:
>\>= 2.6.12: Added the EX, PX, NX and XX options.

(Version 2.6.12 release: [2013-03-29](https://code.google.com/archive/p/redis/downloads))

**Requirements:** Currently the minimum Redis requirement seems to be 2.0.0 (released [2010-09-03](https://code.google.com/archive/p/redis/downloads?page=4)) as `SETEX` is used. However we would reduce the minimum requirement to Redis 1.0.0 using `SETNX` as proposed.

**Support:** phpredis ([doc](https://github.com/phpredis/phpredis#set)) and predis ([somewhat related issue](https://github.com/nrk/predis/issues/309#issuecomment-221203097)) are supporting these options.

**Performance:** From a performance standpoint one command should be ideal.

tillkruss:
This is where it would be ideal to collect telemetry. I’ve never seen anything older than Redis 3, but who know.

naxvog:
Well we could just use `version_compare` to implement this. The performance increase might be worth it. As long as we do the version compare once we should be fine.

ffxluca:
> Just to reiterate:
> ... 
> The fix seems fine to me. Nice catch!
You got it, thanks for rephrasing it!

> However as far as I know both `SETEX` and `SETNX` are atomic redis operations so the comments are a bit misleading. The problem that introduces the race condition is the existence check not the command itself.
Yes! I can remove the inline comments then.

> On the performance side this should perform better than the old implementation as the existence check is done by `SETNX` and the expiry is just set if it is required. So there is a minimum of 1 and a maximum of 2 redis commands sent whereas before there were always 2 being sent while adding a key.

tillkruss:
@ffxluca: Can you push the change from `EXPIRE + SETNX` to use `SET $key $value NX EX $expiration?

tillkruss:
@naxvog: Do you want bring this home, or shall I?

naxvog:
Will push a PR in a bit

PR #175, opened 2020-06-01
added plugin instanciation function

naxvog:


tillkruss:
Mhh, if another plugin defined `redis_object_cache()` the function plugin is never instantiated.

Is the primary purpose of the function to return the instance?

How about we add `RedisObjectCache::init()` and `RedisObjectCache::instance()` methods.

Then `redis_object_cache()` can call for short access and checks `RedisObjectCache::instance()`.

naxvog:
Sure enough, thought I should make the most subtle of changes first but we can transform the whole class to a singleton.

PR #176, opened 2020-06-01
WordPress Coding Standards

naxvog:
Added coding standards XML definition based on WordPress Coding Standards and PHP Compatibility WP using PHP Version 5.4 as well as necessary composer packages as `require-dev`.

Also added exceptions for shorthand array and ternary notation.

_Had some trouble doing an atomic commit so ignore those 2 other commits that I reverted later :wink:_


tillkruss:
Thanks, @naxvog! 

Can you rebase, so I can merge?

PR #182, opened 2020-06-03
Base file cleanup

naxvog:
Moved the class from the main plugin file to its own file and called it plugin.
Added some useful constants to the main file and replaced all referneces of file within the new plugin class.


tillkruss:
Fantastic 👌🏻

PR #183, opened 2020-06-04
Added autoloader

naxvog:
Needed to rename `plugin.php` as it had no `class-` prefix before.

### DIFF
--- a/redis-cache.php
+++ b/redis-cache.php
@@ -26,14 +26,14 @@ $meta = get_file_data( WP_REDIS_FILE, [ 'Version' => 'Version' ] );
 
 define( 'WP_REDIS_VERSION', $meta['Version'] );
 
-if ( defined( 'WP_CLI' ) && WP_CLI ) {
-    require_once dirname( __FILE__ ) . '/includes/wp-cli-commands.php';
-}
-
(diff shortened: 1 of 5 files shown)


tillkruss:
Can you ping when these are ready, or mark them as drafts until they are ready?

naxvog:
I put all currently existing classes into the new namespace so this commit is ready.

redis-cache.php
```diff
@@ -26,12 +26,16 @@
 
 define( 'WP_REDIS_VERSION', $meta['Version'] );
 
+require_once WP_REDIS_PLUGIN_PATH . '/includes/class-autoloader.php';
```
tillkruss, line 29: Is `WP_REDIS_VERSION` still being defined?

naxvog: `WP_REDIS_VERSION` is now defined dynamically to the version definition in the plugin header comment.

redis-cache.php
```diff
@@ -26,12 +26,16 @@
 
 define( 'WP_REDIS_VERSION', $meta['Version'] );
```
tillkruss, line 27: @naxvog: Ich war wohl total blind 🙈

PR #186, opened 2020-06-06
Added Predis as a composer dependency

naxvog:
Additionally separated the `WP_Object_Cache` class from `object-cache.php` to adhere to the coding standards.


includes/class-servers-list.php
```diff
@@ -51,7 +51,7 @@ function ( $value, $key ) use ( &$hidden ) {
 
     public function prepare_items() {
         if ( ! class_exists( 'Predis\Client' ) ) {
-            require_once dirname( __FILE__ ) . '/predis/autoload.php';
+            require_once WP_REDIS_PLUGIN_PATH . '/vendor/predis/predis/autoload.php';
```
tillkruss, line 54: Should we commit the `vendor` folder to the repo with this PR?

includes/class-wp-object-cache.php
```diff
@@ -0,0 +1,1369 @@
+<?php
+
+namespace Rhubarb\RedisCache;
```
tillkruss, line 3: Let's not rename `object-cache.php` to `class-wp-object-cache.php`, because it's just a dropin, not a class part of the plugin.

includes/class-servers-list.php
naxvog: You are right, we obviously need include the `vendor/predis` directory!

includes/class-wp-object-cache.php
naxvog: The drop-in itself is contained in a file but as `WP_Object_Cache` is a class it needs its own class file. We could copy the class file to the `wp-content` folder and include it without referencing the plugin folder.

tillkruss: I rather keep it as is. We can add some PHPCS comments/rules to prevent it from generating a warning 🙈

includes/class-servers-list.php
naxvog: Resolved in https://github.com/rhubarbgroup/redis-cache/pull/186/commits/300d9dc6d9fd370e15c7ac6e688d2983cc879883

includes/class-wp-object-cache.php
naxvog: Actually it does not seem to generate a warning at all...
Resolved in https://github.com/rhubarbgroup/redis-cache/pull/186/commits/300d9dc6d9fd370e15c7ac6e688d2983cc879883

PR #189, opened 2020-06-11
Implement Credis as alternative library to connect to Redis

naxvog:
Test done:
- [x] single instance
- [x] Replication
- [x] Sentinel (only one sentinel instance supported)
- [x] Clustering
- [x] Sharding (not supported)


includes/object-cache.php
```diff
@@ -636,6 +640,116 @@ protected function connectUsingPredis( $parameters ) {
...
+            $args['is_sentinel'] = true;
+        }
+
+        // Credis uses phpredis if it detects it unless we force it to run standalone.
+        $this->redis->forceStandalone();
+
+        $this->redis->connect();
+
+        $this->diagnostics = array_merge(
+            [ 'client' => $client ], // Sadly Credis provides no version information.
```
tillkruss, line 748: If we use fixed versions in the `composer.json` we can parse it and add the version here 👍🏻

includes/object-cache.php
```diff
@@ -636,6 +640,116 @@ protected function connectUsingPredis( $parameters ) {
...
+                        'Credis "%s" library not found. Re-install Redis Cache plugin or delete object-cache.php.',
+                        str_replace( '.php', '', $sub_path )
+                    )
+                );
+            }
+        }
+
+        if ( defined( 'WP_REDIS_SHARDS' ) ) {
+            // TODO: untested!
+            // Sharding not supported?
```
tillkruss, line 696: Maybe we should throw for all unsupported configurations (PhpRedis needs that too). Thoughts?

includes/object-cache.php
```diff
@@ -636,6 +640,116 @@ protected function connectUsingPredis( $parameters ) {
...
+
+        // Load bundled Credis library.
+        if ( ! class_exists( 'Credis_Client' ) ) {
+            $to_load[] = 'Client.php';
+        }
+
+        if ( ( defined( 'WP_REDIS_SHARDS' ) || defined( 'WP_REDIS_SENTINEL' ) || defined( 'WP_REDIS_SERVERS' ) || defined( 'WP_REDIS_CLUSTER' ) ) && ! class_exists( 'Credis_Cluster' ) ) {
+            $to_load[] = 'Cluster.php';
+
+            if ( true
```
tillkruss, line 672: Debug?

includes/object-cache.php
```diff
@@ -385,7 +385,11 @@ public function __construct( $fail_gracefully = true ) {
                 case 'phpredis':
                     $this->connectUsingPhpRedis( $parameters );
                     break;
+                case 'credis':
+                    $this->connectUsingCredis( $parameters );
+                    break;
```
tillkruss, line 390: Should we default to Credis in v2, or leave the switch for v2.1.0?

includes/object-cache.php
naxvog: A configuration checker class would be ideal. I almost started writing one a couple of days ago :wink:.
If we allow configuration via UI this would be necessary.

includes/object-cache.php
naxvog: Actually I do write all conditions with more than one condition using this approach as it allows for fast order switching or commenting out. Also I personally am a line 80 guy :wink:. The compiler removes the first boolean anyway. 
Forgot to transform it back, will update in a couple if hours.

includes/object-cache.php
naxvog: Well if we get all connection parameters tested we could switch but it would be better to give the user the option to test `credits` first (and potentially report problems).

includes/object-cache.php
naxvog: Thought about that but it seems like a quite hacky approach. Other than opening an issue on Credits itself I don't see a better one.

tillkruss: Let's do that, I'll open an issue and see how fast they are.

`Latest commit on Nov 26, 2019` yikes...

includes/object-cache.php
naxvog: changed in https://github.com/rhubarbgroup/redis-cache/pull/189/commits/f03b1572e3f78679ab27b062b5cde40ba4d2a9e5

includes/object-cache.php
tillkruss: Waiting for merge: https://github.com/colinmollenhour/credis/pull/136

tillkruss: Merged: https://github.com/colinmollenhour/credis/pull/136

naxvog: Resolved in ff16ed915310e05fb0908cc6e50b43a29d410b7f

PR #190, opened 2020-06-11
Exists+set race condition

naxvog:
Updated https://github.com/rhubarbgroup/redis-cache/pull/170 to include Redis version checks.
Now every `add_or_replace` method call should result in one Redis operation - preventing any race condition.
Added private property `redis_supports` to keep track of features that can be used to avoid excessive `version_compare` function calls.

Note on compatibility:
`Credis` supports the `phpredis` `set` method call using 3 parameters


tillkruss:
Mhhh. Redis 2.6.12 was released on Mar 29, 2013. If someone is running that they can stick with v1 of the plugin.

tillkruss:
Closes #170 and #166.

ffxluca:
Thanks guys!

PR #197, opened 2020-06-19
fix support database selection with sentinel

bertuxdeveloper:
HI @tillkruss,
we use Redis with sentinel on our infrastructure and I notice that the DB selected is already the 0 even if we have set the WP_REDIS_DATABASE with a different value.

I see that predis library support the database choice https://github.com/nrk/predis/issues/291# and I added the $options['parameters']['database'] = $parameters['database']; in the same way of WP_REDIS_PASSWORD and it works.

Thanks in advance.
Fabio


tillkruss:
Thank you!

PR #204, opened 2020-06-28
Admin UI advertising card now adheres to DISABLE_BANNERS constant

naxvog:
The [Wiki](https://github.com/rhubarbgroup/redis-cache/wiki/Configuration-Options#wp_redis_disable_banners-default-not-set) states that the `WP_REDIS_DISABLE_BANNERS` constant should disable all promotions for Redis Cache Pro but the Admin UI currently has the corresponding card displayed at all times.

### DIFF
--- a/includes/ui/admin-page.php
+++ b/includes/ui/admin-page.php
@@ -221,6 +221,8 @@
 
     </div>
 
+    <?php if ( ! defined( 'WP_REDIS_DISABLE_BANNERS' ) || ! WP_REDIS_DISABLE_BANNERS ) : ?>
+
     <div class="pro-version">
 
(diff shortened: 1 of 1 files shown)


tillkruss:
I removed that intentionally in v2. My reasoning was that banners make sense to disable, but the Pro card inside the settings is separate. I'd be open to introduce another constant for it, or leave it as is, like most freemium plugins do.

tillkruss:
As discussed via Slack: We keep the settings screen as is, but `WP_REDIS_DISABLE_BANNERS` will remove all banners and admin notices. 

See #205 for removing Pro from the dashboard widget.

PR #211, opened 2020-07-03
JS refactoring to increase compatibility and reduce size

naxvog:
- added jQuery event namespacing `click.redis`
- added the ability to use a non WP-standard jQuery version
- used jQuery.extend to extend the chart defaults while creating the different charts
- mostly rely on one object in the global context
- moved the UI-facing function executions to the bottom
- wrapped all in a closure to avoid conflicts

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -176,15 +176,18 @@ class Plugin {
 
         wp_localize_script(
             'redis-cache',
-            'rediscache_l10n',
+            'rediscache',
             array(
-                'time' => __( 'Time', 'redis-cache' ),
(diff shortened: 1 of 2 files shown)


includes/class-plugin.php
```diff
@@ -699,4 +702,11 @@ public function on_deactivation( $plugin ) {
 
         ob_end_clean();
     }
+
+    public function action_link( $action ) {
```
tillkruss, line 706: @naxvog: What's this for?

naxvog: The method should be part of another PR will revert it (for now 😉)

PR #215, opened 2020-07-05
OSX compat constant line deletion

naxvog:
Might fix the currently broken method of deleting a line with a given constant name in it to load another docker configuration

### DIFF
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -27,6 +27,7 @@ function stop {
 # Modifies the auto-prepend-file
 function apf {
     echo "APF-Constant $@"
+    ESCAPE='s/[]\/$*.^[]/\\&/g'
     # --reset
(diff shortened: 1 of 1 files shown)


tillkruss:
```
<?php

$constants = [
    // constant-definition start
    'WP_REDIS_CLIENT' => 'predis',
    'WP_REDIS_SERVERS' => ['tcp://172.20.0.2:6379?alias=master','tcp://172.20.0.6:6379?alias=slave-01','tcp://172.20.0.4:6379?alias=slave-02','tcp://172.20.0.5:6379?alias=slave-03',],
    // constant-definition end
];

foreach ( $constants as $constant => $value ) {
    if ( ! defined( $constant ) ) {
        define( $constant, $value );
    }
}
```

PR #216, opened 2020-07-07
General cleanup

naxvog:
Some changes to shed some unnecessary stuff and fix some minor errors.


tillkruss:
@naxvog: ping me when this is ready.

assets/css/admin.css
```diff
@@ -1,4 +1,3 @@
-
```
tillkruss, line 1: Can we turn this annoying rule off?

includes/class-qm-output.php
```diff
@@ -45,8 +50,8 @@ public function admin_menu( array $menu ) {
 
     public function panel_menu( array $menu ) {
         $ids = array_keys( $menu );
-        $request = array_search( 'qm-request', $ids );
-        $position = $request === false ? count( $menu ) : $request;
+        $request = array_search( 'qm-request', $ids, true );
+        $position = false === $request ? count( $menu ) : $request;
```
tillkruss, line 54: 🤮

includes/ui/diagnostics.php
```diff
@@ -74,42 +80,43 @@
...
+        $info[ $constant ] = wp_json_encode( constant( $constant ) );
     }
 }
 
 if ( defined( 'WP_REDIS_PASSWORD' ) ) {
     $password = WP_REDIS_PASSWORD;
 
     if ( is_array( $password ) ) {
-        if ( isset( $password[1] ) && ! is_null( $password[1] ) && $password[1] !== '' ) {
+        if ( isset( $password[1] ) && ! empty( $password[1] ) ) {
```
tillkruss, line 91: This has to be reverted, since `0` is empty.

includes/ui/diagnostics.php
```diff
@@ -74,42 +80,43 @@
...
             $password[1] = str_repeat( '•', 8 );
         }
 
-        $info[ 'WP_REDIS_PASSWORD' ] = json_encode( $password );
-    } elseif ( ! is_null( $password ) && $password !== '' ) {
```
tillkruss, line 90: Same here.

redis-cache.php
```diff
@@ -13,9 +13,11 @@
  * GitHub Plugin URI: https://github.com/rhubarbgroup/redis-cache
  * License: GPLv3
  * License URI: http://www.gnu.org/licenses/gpl-3.0.html
+ *
+ * @package RhubarbGroup/RedisCache
  */
 
-defined( 'ABSPATH' ) || exit;
+defined( 'ABSPATH' ) || exit( 'direct access forbidden' );
```
tillkruss, line 20: Can we stick to just an exit without a message?

includes/ui/settings.php
```diff
@@ -93,19 +98,31 @@ class="section section-<?php echo esc_attr( $tab->slug ); ?> <?php echo $tab->de
...
+                                    );
+                                ?>
                             </li>
                         <?php endif; ?>
 
-                        <?php if ( ! $phpRedisInstalled ) : ?>
+                        <?php if ( ! $phpredis_installed ) : ?>
                             <li>
-                                <?php printf( esc_html__( 'The PhpRedis extension is not installed.', 'redis-cache' ), phpversion() ); ?>
+                                <?php printf( esc_html__( 'The PhpRedis extension is not installed.', 'redis-cache' ), esc_html( phpversion() ) ); ?>
```
tillkruss, line 115: We can remove the `printf()` eh?

includes/ui/settings.php
```diff
@@ -93,19 +98,31 @@ class="section section-<?php echo esc_attr( $tab->slug ); ?> <?php echo $tab->de
...
                     <ul>
-                        <?php if ( ! $isPhp7 ) : ?>
+                        <?php if ( ! $is_php7 ) : ?>
                             <li>
-                                <?php printf( esc_html__( 'The current version of PHP (%s) is too old. PHP 7.0 or newer is required.', 'redis-cache' ), phpversion() ); ?>
+                                <?php
+                                    printf(
+                                        // translators: %s = PHP Version.
+                                        esc_html__( 'The current version of PHP (%s) is too old. PHP 7.0 or newer is required.', 'redis-cache' ),
+                                        esc_html( phpversion() )
```
tillkruss, line 107: Isn't this double encoded?

includes/ui/tabs/overview.php
```diff
@@ -169,41 +175,71 @@
...
         <tr>
             <th><?php esc_html_e( 'Read Timeout:', 'redis-cache' ); ?></th>
-            <td><code><?php echo sprintf( esc_html__( '%ss', 'redis-cache' ), $diagnostics['read_timeout'] ); ?></code></td>
+            <td>
+                <code>
+                    <?php
+                        echo sprintf(
+                            // translators: %s = Redis read timeout in seconds.
+                            esc_html__( '%ss', 'redis-cache' ),
+                            esc_html( $diagnostics['read_timeout'] )
```
tillkruss, line 201: double encoded?

includes/ui/tabs/overview.php
```diff
@@ -169,41 +175,71 @@
...
         <tr>
             <th><?php esc_html_e( 'Retry Interval:', 'redis-cache' ); ?></th>
-            <td><code><?php echo sprintf( esc_html__( '%sms', 'redis-cache' ), $diagnostics['retry_interval'] ); ?></code></td>
+            <td>
+                <code>
+                    <?php
+                        echo sprintf(
+                            // translators: %s = Redis retry interval in milliseconds.
+                            esc_html__( '%sms', 'redis-cache' ),
+                            esc_html( $diagnostics['retry_interval'] )
```
tillkruss, line 218: double encoded?

redis-cache.php
naxvog: Relic of my testing - removed

includes/ui/settings.php
naxvog: Why would it be?
First the translated string is escaped including the `%s` and then the escaped `phpversion` call is replacing the placeholder.

However we could do something like this to save one `esc_html` call:
```php
echo esc_html(
  sprintf(
    // translators: %s = PHP Version.
    __( 'The current version of PHP (%s) is too old. PHP 7.0 or newer is required.', 'redis-cache' ),
    phpversion()
  )
);
```

includes/ui/settings.php
naxvog: Absolutely right!
If you do too many of these changes you are going to get blind eventually :wink:

assets/css/admin.css
naxvog: Not sure if I follow: the empty line was there on purpose?
We can get rid of the rule but it might affect non `*.css` files aswell. Need to look into it.

includes/class-qm-output.php
naxvog: Well yoda style conditions do indeed reduce hastily made mistakes but I guess this type of error (forgetting one "=") is mostly done by beginners. Personally I changed my coding style to incorporate them as I think that this is a quite clever trick :wink:

Do you think parentheses would reduce your aversion? 😁

tillkruss: They are ok 👍🏻 I just don’t like them personally.

Issue #217, opened 2020-07-09
Combine composer.json files


naxvog:
Just used this on another project and thought that it might be beneficial to use here too:
https://github.com/mnsami/composer-custom-directory-installer

Might push a PR as soon as I finish work :wink:

tillkruss:
Very cool

tillkruss:
Closed via #218.

PR #218, opened 2020-07-09
Removed additional composer.json

naxvog:
As mentioned in #217 I combined both composer.json files using [`mnsami/composer-custom-directory-installer`](https://github.com/mnsami/composer-custom-directory-installer).


tillkruss:
Should we collapse the path from `/dependencies/vendor/package` to `/dependencies/package`?

naxvog:
Thought about it but wanted to be as lean as possible bu as you mentioned it; give me a second.

PR #231, opened 2020-07-18
Fix support for custom ports with credis

markhughes:
I changed to credis today, and it turns out the connection string needs to also provide the port here.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -761,7 +761,7 @@ class WP_Object_Cache {
                 : $parameters['host'];
 
             $args = [
-                "{$parameters['scheme']}://{$host}",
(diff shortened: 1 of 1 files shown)


tillkruss:
Thanks, I'll push this into v2.0.5 tomorrow.

tillkruss:
@markhughes: I can't reproduce this. What error are you seeing? Can you post your diagnostics without this fix?

tillkruss:
@markhughes: Can you try to use the current `develop` branch and see if that fixes your connection issues?

markhughes:
Looks like this resolved it: https://github.com/rhubarbgroup/redis-cache/commit/b10277edc64c1ef7f063614522381eb49d907dd3#diff-3ea3a8942f84c506446c32d498ac655dR759-R764

Thanks!

Issue #234, opened 2020-07-22
Switching tabs unexpected anchor scrolling


naxvog:
With the removal of the `#top` anchor before the tab target (e.g. `#top#metrics`) switching a tab automatically scrolls to the tab content. This results in the tab bar itself not showing and is quite frankly an awkward user experience.

The `#top` anchor was removed because forced scrolling should be avoided to improved the UX but the current behaviour is worse in my opinion.
Clicking a tab should always result in displaying the top part of the page including the tab bar.

The original approach was modelled after Yoast SEO. Given its number of installations, the plugin's scope and its rapid development cycle I would assume that they did some UX testing and concluded that this behaviour is superior to others.

tillkruss:
Can you post a screen recording of the issue?

naxvog:
![screencast-20200723-09-07-34-Tri](https://user-images.githubusercontent.com/4963183/88260588-f851fb00-ccc4-11ea-9196-bc1623f3512f.gif)

tillkruss:
Wow, is that IE?

naxvog:
FF 79.0b9

naxvog:
But this should be the default behaviour in all browsers as the anchor of the content element is selected. Returning `false` in the jQuery method should result in a fix as the click event won't propagate further.

tillkruss:
@naxvog: Feel free to improve 88e9577 if necessary.

naxvog:
in 2.0.7 tabs are unusable:
Tab nav links don't have an anchor linking to `wp-admin/metrics` for example. Tabs are not hidden by default. Its a mess currently.

Please consider implementing my tested PR #235 

![DOM-Node screenshot](https://user-images.githubusercontent.com/4963183/88772466-1dd97b80-d181-11ea-8ff9-36d3cced3e7f.png)

tillkruss:
The `#top` solution is backwards. The `admin.js` file wasn't properly copied to wp.org's SVN repo.

PR #235, opened 2020-07-23
improved tab switching using js history api

naxvog:
Improved upon #234

History is now correctly set and respected on back and forward navigation or jumps.

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -406,32 +406,61 @@
     $(function () {
         var $tabs = $( '#redis-tabs' );
 
-        $tabs.find( 'a' ).on(
-            'click.redis',
-            function ( event ) {
-                event.preventDefault();
+        if ( $tabs.length ) {
 
-                var $this = $( this );
-                var $target = $( $this.data( 'target' ) );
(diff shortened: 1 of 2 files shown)


tillkruss:
Closing in favor of 7bb1f3e.

tillkruss:
@naxvog: Do you still have some kind of jumping occur when navigating or clicking tabs?

naxvog:
Have not tested this but I see the following issues by looking at the code:
* `e.preventDetault` should be `return false` for jQuery events. Either way this should fix the jumping issue.
* `addEventListener` should be specified via jQuery and namespaces
* History API Implementation is worthless if you only push states but don't react to those states if navigation occurs (back/forward)
* The `#top` anchor needs to exist to ensure that a link to a specific tab (from another page) will still show the top of the page and not just the tab.

tillkruss:
>History API Implementation is worthless if you only push states but don't react to those states if navigation occurs (back/forward)

That's working with `hashchange`, thanks for pointing it out.

>The #top anchor needs to exist to ensure that a link to a specific tab (from another page) will still show the top of the page and not just the tab.

That's because of bad design. Will fix that

Issue #240, opened 2020-07-29
Flush button not showing up


xberg:
Hi,
I upgraded today from 1.x to 2.0.7 and I am having the following problems:
- no more flush / disable button as in version 1.x
- the Metrics and Diagnostics tabs do nothing when clicked

This is what I see in overview:

Status: Connected
Client: PhpRedis (v5.3.0)
Drop-in: Valid
Disabled: No
Filesystem: Working
Ping: 1
Errors: []
PhpRedis: 5.3.0
Predis: Not loaded
Credis: Not loaded
PHP Version: 7.3.20
Plugin Version: 2.0.7
Redis Version: 2.8.17

tillkruss:
Relevant: https://wordpress.org/support/topic/no-flush-button-metrics-and-diagnostics-go-nowhere/#post-13186587

tillkruss:
@xberg Are you seeing any errors in your PHP log files?

Is it a multisite installation?

tillkruss:
Try updating to 2.0.8 and:

- Delete the `object-cache.php` dropin by hand (FTP)
- Flush you Redis database (`redis-cli flushall`)
- Re-enable the plugin via `Settings -> Redis`
- Try flushing again

naxvog:
The Flush button currently only shows up if a redis connection is established successfully.

Related: #241

tillkruss:
I'll also test this with a foreign dropin and when the dropin can't be read, maybe an invalid header or something.

tillkruss:
@xberg: Can we debug this via Zoom? This seems like an obscure bug.

xberg:
Hi,
I upgraded to 2.0.10 and everything works perfectly now. Well done!
sorry for my late reply!

PR #241, opened 2020-07-29
Flush cache button now always shown and disabled if necessary

naxvog:


tillkruss:
I'm not sure how this would solve that persons issue. Seems like an obscure bug to me and displaying the button doesn't resolve it, it would just be disabled until the issue is resolved.

There is some issue with `get_status()` (connected) and `get_redis_status()` (null) not lining up.

naxvog:
This PR is only to be more consistent as there are always 2 buttons shown. It was never meant as a solution to #240.
A button missing might be an error, a button disabled is hinting at an issue with the configuration.

tillkruss:
Yeah, I've refactored the status logic to hopefully show these weird corner cases in b2cd679.

PR #242, opened 2020-07-29
Docker improvements

naxvog:
You can now easily define the client to be used using the `-c` or the `--client` flag.
Example: `docker/start.sh -c credis replication`

Will update the wiki once this is merged.


naxvog:
Updated the Wiki pages accordingly

PR #246, opened 2020-08-15
Phpcs improvements

naxvog:
On the road to 0 `phpcs` problems :wink:

- [x] `cli/class-commands.php`
- [x] `ui/tabs/diagnostics.php`
- [x] `ui/tabs/metrics.php`
- [x] `ui/tabs/overview.php`
- [x] `ui/class-tab.php` (new)
- [x] `ui/diagnostics.php`
- [x] `ui/query-monitor.php`
- [x] `ui/settings.php`
- [x] `ui/widget.php`
- [x] `includes/class-autoloader.php`
- [x] `includes/class-plugin.php`
- [x] `includes/class-qm-collector.php`
- [x] `includes/class-qm-output.php`
- [x] `includes/class-ui.php`
- [x] `includes/object-cache.php`
- [x] `redis-cache.php`


includes/class-plugin.php
```diff
@@ -747,7 +916,7 @@ public function maybe_print_comment() {
         );
 
         if ( ! WP_DEBUG ) {
-            printf( "\n<!-- %s -->\n", $message );
+            printf( "\n<!-- %s -->\n", esc_html( $message ) );
```
tillkruss, line 919: It's a comment, not HTML 🙂

naxvog: Any suggestion for another output sanitization function? In case we decide to translate the comment `esc_html` would be needed.

includes/ui/tabs/overview.php
```diff
@@ -7,14 +7,12 @@
 
 defined( '\\ABSPATH' ) || exit;
 
-$plugin = redis_object_cache();
```
tillkruss, line 10: Why is this `$roc` and not `$plugin`?

includes/ui/tabs/overview.php
```diff
@@ -161,7 +160,12 @@
         <tr>
             <th><?php esc_html_e( 'Password:', 'redis-cache' ); ?></th>
             <td>
-                <code><?php echo str_repeat( '&#8226;', 8 ); ?></code>
+                <code>
+                    <?php
+                        // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+                        echo str_repeat( '&#8226;', 8 );
+                    ?>
+                </code>
```
tillkruss, line 168: Should we just do the actual text without `str_repeat()`?

includes/object-cache.php
```diff
@@ -14,7 +14,9 @@
 
 defined( '\\ABSPATH' ) || exit;
 
-if ( ! defined( 'WP_REDIS_DISABLED' ) || ! WP_REDIS_DISABLED ) :
+if ( ! defined( 'WP_REDIS_DISABLED' ) || ! WP_REDIS_DISABLED ) {
+    return;
+}
```
tillkruss, line 19: This doesn't work, try disabling the object cache with it.

includes/object-cache.php
```diff
@@ -498,7 +505,7 @@ protected function build_parameters() {
         if ( isset( $parameters['password'] ) ) {
             $password = $parameters['password'];
 
-            if ( is_null( $password ) || $password === '' ) {
+            if ( is_null( $password ) || '' === $password ) {
```
tillkruss, line 508: Can we turn this Yoda stuff off? I don't like it, even if it is WP's recommendation.

includes/class-plugin.php
tillkruss: We won't translate the output, since it's used by services to identify the plugin.

includes/class-plugin.php
```diff
@@ -747,12 +925,13 @@ public function maybe_print_comment() {
...
 
             return;
         }
 
-        $bytes = strlen( serialize( $wp_object_cache->cache ) );
+        // TODO: find a better method to determine cache size as using `serialize` is discouraged.
+        $bytes = strlen( serialize( $wp_object_cache->cache ) ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
```
tillkruss, line 934: `json_encode()`?

includes/class-plugin.php
```diff
@@ -762,15 +941,23 @@ function_exists( 'size_format' ) ? size_format( $bytes ) : "{$bytes} bytes",
             $wp_object_cache->diagnostics['client']
         );
 
-        printf( "<!--\n%s\n\n%s\n-->\n", $message, $debug );
+        printf( "<!--\n%s\n\n%s\n-->\n", esc_html( $message ), esc_html( $debug ) );
```
tillkruss, line 944: We can drop this `esc_html()` as well.

includes/ui/tabs/overview.php
naxvog: `$plugin` is considered a global WordPress variable and should therefore not be overwritten. As we work with namespaces this should not be a problem but I guess it is best practice to avoid naming variables after WordPress globals.

includes/object-cache.php
naxvog: 🤦 Well it would work if I had actually done the necessary boolean conversion to the negative form. Tested that and it worked. Will push the new one in a second.

includes/object-cache.php
```diff
@@ -1587,6 +1587,7 @@ public function decrement( $key, $offset = 1, $group = 'default' ) {
     /**
      * Render data about current cache requests
      *
+     * @deprecated No longer used by the plugin.
```
tillkruss, line 1590: This is actually being called by the Debug Bar plugin.

includes/object-cache.php
tillkruss: Are you 100% sure that the `wp_cache_*()` functions are not being declared this way anyhow? I recall reverting this change once before.

naxvog: Functions will be defined before the code itself is run so returning early will still define the cache functions. Reverted the commit.

tillkruss:
Can you undo the indentation? No need for it.

includes/ui/diagnostics.php
```diff
@@ -140,6 +140,6 @@
     if ( defined( 'WP_CLI' ) && WP_CLI ) {
         WP_CLI::line( "{$name}: $value" );
     } else {
-        echo "{$name}: {$value}\r\n";
+        echo esc_html( "{$name}: {$value}\r\n" );
```
tillkruss, line 143: Das sollte `esc_textarea()` sein, oder?

includes/object-cache.php
```diff
@@ -837,8 +836,8 @@ protected function connectUsingHHVM( $parameters ) {
      */
     public function fetch_info() {
         $options = method_exists( $this->redis, 'getOptions' )
-            ? $this->redis->getOptions()
-            : new stdClass();
+        ? $this->redis->getOptions()
+        : new stdClass();
```
tillkruss, line 840: There are a couple of places where the indentation is removed, can we turn that rule off and undo it?

includes/object-cache.php
```diff
@@ -930,8 +929,8 @@ public function replace( $key, $value, $group = 'default', $expiration = 0 ) {
      */
     protected function add_or_replace( $add, $key, $value, $group = 'default', $expiration = 0 ) {
         $cache_addition_suspended = function_exists( 'wp_suspend_cache_addition' )
-            ? wp_suspend_cache_addition()
-            : false;
+        ? wp_suspend_cache_addition()
+        : false;
```
tillkruss, line 933: Indentation.

tillkruss:
Notes to myself.

- [ ] Test dismissing banners
- [ ] Test diagnostics
- [ ] Fix up `connect_using_credis()`
- [ ] Fix up `fetch_info()`
- [ ] Fix up `function_exists( 'wp_suspend_cache_addition' )`
- [ ] Fix up `return function () use ( $salt )`

includes/object-cache.php
naxvog: Seems like this happens during `phpcbf`. Need to keep an eye out for this. There is no particular rule for this as far as I can see.

PR #256, opened 2020-08-21
Cron missed logic

naxvog:
Adds logic to detect cron events failing to execute (event did not run within 15 minutes) and adds an admin notice if this is the case.

The notice might need a link to some troubleshooting article.


tillkruss:
I think this isn't needed. Let's resurrect it a 2nd person has the issue 🙂

Issue #257, opened 2020-08-24
Pass original expiry to redis_cache_expiration


markhughes:
I have a case where I want some key types to override the original `WP_REDIS_MAXTTL` being used..

The filter `redis_cache_expiration` does not pass the original expiry, I'm currently hacking it in manually by changing:

```php
$expiration = apply_filters( 'redis_cache_expiration', $this->validate_expiration( $expiration ), $key, $group );
```

to:

```php
$expiration = apply_filters( 'redis_cache_expiration', $this->validate_expiration( $expiration ), $key, $group, $expiration );
```

However, you might know an alternative to this, I had a few ideas such as a adding a `should_validate_expiration` for key/group filter. What are your thoughts? 

If you think my current solution is good, or a `should_validate_expiration` filter I can go ahead and create a PR for you. :)

tillkruss:
You mean the 1st argument is always overwritten by `WP_REDIS_MAXTTL`? If so, then yeah, let's pass a 4th argument to that filter 👍🏻

```php
$expiration = apply_filters(
    'redis_cache_expiration',
    $this->validate_expiration( $expiration ), // <-- is overwritten by `WP_REDIS_MAXTTL`
    $key,
    $group
);
```

markhughes:
Cheers, redis-cache is the only redis cache plugin that really adds in the extra "ooft" that developers need 😄

tillkruss:
Thanks Mark! If you want to leave a review on WP.org, that would be magnificent 😗

PR #258, opened 2020-08-24
Hook comments

naxvog:
- Documented all hooks
- Fixed a wrong type hint
- Incorporated suggestion from #257


naxvog:
Created https://github.com/rhubarbgroup/redis-cache/wiki/Hook-reference as well

PR #262, opened 2020-08-30
Changed all array definitions to shorthand style

naxvog:
Let me know if we should use the shorthand `[ 1, 2, 3 ]` (this PR) or the old `array( 1, 2, 3 )`.
Either way we should consistently use one of them.


tillkruss:
Thanks, I've been meaning to do that for v2.x

Issue #259, opened 2020-08-24
Improvements


tillkruss:
https://twitter.com/markjaquith/status/1297690855017521153

https://github.com/coenjacobs/mozart

naxvog:
- **No global variables. Ever.**
We need to refactor our templates as they rely on the current instance `$roc`. Would coincide with **Static class first** as we can switch a lot of methods to be static. Obviously we need to rely on `global $wp_object_cache;`
- **Namespace everything.**
Not sure I agree for constants as well as the plugin entry file. Don't know if he would agree to this exceptions. Cache functions can not be namespaced for obvious reasons. Namespacing the `WP_Object_Cache` class requires #251 or similar.
- **Static class first**
We can and should switch all methods no requiring state to static ones. Especially `WP_Object_Cache` might profit from the increased performance not having to load the current instance on certain methods.
- **Avoid the god object**
We should definitely break the `Plugin` class into logical parts. Might need to break apart `WP_Object_Cache` too.
- **Defer init to `plugins_loaded`** / **One hook init**
I was under the impression that I opened a PR for this a while ago but it apparently didn't make it into master. Would be quite easy to do using the following code (ignores the "don't create instances" part)
```php
add_action( 'plugins_loaded', Rhubarb\RedisCache\Plugin::class, 'instance' );
```
- **Mozart**
Did try to use Mozart in the past but couldn't figure it out completely. Documentation is lackluster imho. If we can get it configured it would be quite nice to have though.

PR #263, opened 2020-08-30
"Statification" of the plugin class

naxvog:
Transformed every method of the plugin class to a static one. Started from the bottom up and replaced all calls as needed.

Introduced getter methods for `Plugin::$screen` and `Plugin::$page` and moved the constructor logic to a new `init` method - scrapped the singleton pattern in the process.

Removed the `redis_object_cache` function as there is no more instance to retrieve.

Changes will satisfy #259 No. 4 and 7 for the plugin class.

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -438,7 +438,7 @@ class Plugin {
      *
      * @return string
      */
-    public function get_status() {
+    public static function get_status() {
         global $wp_object_cache;
 
         if ( defined( 'WP_REDIS_DISABLED' ) && WP_REDIS_DISABLED ) {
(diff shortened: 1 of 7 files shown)


naxvog:
signed commits and rebased against [2.0.13](https://github.com/rhubarbgroup/redis-cache/tree/2.0.13)

tillkruss:
@naxvog: This is looking pretty good on a quick glance. How are you feeling about the change?

naxvog:
Looks good from my side. Just tested the CLI commands to be sure :wink:

naxvog:
Rebased to `2.0.15`

PR #268, opened 2020-09-05
Fixed an error due to variable referencing

naxvog:
Removed variable referencing in `connect_using_predis` method as it would cause fatals due to repeating referenced variable name `$connection_string` in consecutive `foreach` loops.


tillkruss:
Are the cluster nodes / replicated servers still listed in the admin screen after this change?

naxvog:
Yes they will. Example of the replication docker preset:
Annotation 2020-09-06 000513

PR #273, opened 2020-09-10
Cache key map

naxvog:
Inspired by #270 I have added a cache key map to avoiding consecutive builds of already built redis cache keys.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1778,13 +1778,15 @@ LUA;
      * @return  string
      */
     public function build_key( $key, $group = 'default' ) {
+        static $keys = [];
+
         if ( empty( $group ) ) {
             $group = 'default';
         }
 
         $key_index = "{$key}:{$group}";
 
-        if ( ! isset( $this->keys[ $key_index ] ) ) {
+        if ( ! isset( $keys[ $key_index ] ) ) {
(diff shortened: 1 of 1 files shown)


includes/object-cache.php
```diff
@@ -284,6 +284,13 @@ class WP_Object_Cache {
      */
     public $cache = [];
 
+    /**
+     * Holds cache key to redis key map
+     *
+     * @var array
+     */
+    public $keys = [];
```
tillkruss, line 292: Since it's centrally managed, should we use a `static $keys = [];` inside the method itself?

tillkruss:
Great work 👍🏻

ocean90:
Just noting that this change broke the use of the drop-in in Multisite. Applying fbfc86fc22b4df0c636b05e5ce5da0b548fd5b08 and flushing the cache will fix it.

naxvog:
```php
$prefix = $this->is_global_group( $group ) ? $this->global_prefix : $this->blog_prefix;
```
`blog_prefix` contains `blog_id` that's why the error occurs - need to consider `$prefix` in `$key_index`. Will post a fixed version shortly.

tillkruss:
@ocean90: I tagged 2.0.15 👍🏻

Issue #271, opened 2020-09-09
Metrics API endpoint


tillkruss:
An upcoming partner rolling out the plugin would love an API endpoint to get Redis Metrics as JSON.

Optionally a WP CLI command as well.

tillkruss:
Object Cache Pro supports this already (CLI and REST).

PR #275, opened 2020-09-11
Metrics custom max time

naxvog:
Introduced `WP_REDIS_METRICS_MAX_TIME` constant to define the maximum time metrics should be collected. Default value is 1 hour. The dashboard widget always shows half an hour due to its size restriction.

This might be useful for external metrics collection as well (related: #271)

Constant so far not documented.

### DIFF
--- a/redis-cache.php
+++ b/redis-cache.php
@@ -28,8 +28,6 @@ $meta = get_file_data( WP_REDIS_FILE, [ 'Version' => 'Version' ] );
 
 define( 'WP_REDIS_VERSION', $meta['Version'] );
 
-defined( 'WP_REDIS_METRICS_MAX_TIME' ) || define( 'WP_REDIS_METRICS_MAX_TIME', HOUR_IN_SECONDS );
-
 require_once WP_REDIS_PLUGIN_PATH . '/includes/class-autoloader.php';
 
 $autoloader = new Rhubarb\RedisCache\Autoloader();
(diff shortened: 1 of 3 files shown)


redis-cache.php
```diff
@@ -28,6 +28,8 @@
 
 define( 'WP_REDIS_VERSION', $meta['Version'] );
 
+defined( 'WP_REDIS_METRICS_MAX_TIME' ) || define( 'WP_REDIS_METRICS_MAX_TIME', HOUR_IN_SECONDS );
```
tillkruss, line 31: We're not defining default constants for others like `WP_REDIS_MAXTTL`. Can we just do a check in `class-plugin.php`?

tillkruss (changes requested):

redis-cache.php
naxvog: introduced a static getter method instead

tillkruss:
Is the constant designed to allow collecting longer periods of data, or to display more data in the settings?

We should probably limit the timespan in the settings chars, too. How does the dashboard chart look when displaying 6 hours of data on a narrow browser window?

naxvog:
The dashboard chart is limited to half an hour regardless of `WP_REDIS_METRICS_MAX_TIME`. The wider chart in the settings page will display all data until `WP_REDIS_METRICS_MAX_TIME `.

The intention is to monitor more data over a longer period of time. I would assume most users want to know how their installation held up in the past 24h for example.

tillkruss:
Alright, let me know 3600 entries look in the chart, especially on smaller screens.

naxvog:
Fixed a crucial typo 😅 

... also screenshots:

Tablet in portrait mode:
metrics-tablet

Mobile phone in portrait mode:
metrics-mobile

PR #276, opened 2020-09-11
Pro version ad translation

naxvog:
As per https://wordpress.org/support/topic/redis-cache-pro-box-not-translatable/

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -298,7 +298,7 @@ class Plugin {
                     'calls' => __( 'Calls', 'redis-cache' ),
                     'no_data' => __( 'Not enough data collected, yet.', 'redis-cache' ),
                     'no_cache' => __( 'Enable object cache to collect data.', 'redis-cache' ),
-                    'pro' => 'Object Cache Pro',
+                    'pro' => __( 'Object Cache Pro', 'redis-cache' ),
                 ],
             ]
         );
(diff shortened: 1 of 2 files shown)


includes/class-plugin.php
```diff
@@ -298,7 +298,7 @@ public function enqueue_admin_scripts() {
                     'calls' => __( 'Calls', 'redis-cache' ),
                     'no_data' => __( 'Not enough data collected, yet.', 'redis-cache' ),
                     'no_cache' => __( 'Enable object cache to collect data.', 'redis-cache' ),
-                    'pro' => 'Object Cache Pro',
+                    'pro' => __( 'Object Cache Pro', 'redis-cache' ),
```
tillkruss, line 301: The name is fixed and doesn't need to be translated.

includes/ui/settings.php
```diff
@@ -74,25 +74,25 @@ class="<?php echo esc_attr( $ui_tab->classes() ); ?>"
 
                 <div class="card">
                     <h2 class="title">
-                        Object Cache Pro
+                        <?php esc_html_e( 'Object Cache Pro', 'redis-cache' ); ?>
```
tillkruss, line 77: The name is fixed and doesn't need to be translated.

tillkruss (changes requested):

tillkruss:
I'm not stoked about giving translators the ability to write ad copy.

PR #277, opened 2020-09-11
Investigate weird serialization logic

tillkruss:
My best guess is that this is never called. Thoughts @naxvog?

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -2011,7 +2011,7 @@ LUA;
             return serialize( $data );
         }
 
-        if ( $this->is_serialized( $data, false ) ) {
+        if ( ! $this->is_serialized( $data, false ) ) {
             // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
             return serialize( $data );
         }


naxvog:
Just to summarize our slack conversation:

- The original change was made years ago
- The last condition of `maybe_serialize` serializes an already serialized string (nonsense)
- All non strings should be serialized

We agreed on taking no further action at the moment in 2.x but will refactor the method in a future 3.0 release in order to maintain compatibility.

includes/object-cache.php
```diff
@@ -2006,17 +2006,12 @@ protected function maybe_serialize( $data ) {
...
         if ( $this->is_serialized( $data, false ) ) {
-            // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
-            return serialize( $data );
+            return $data;
         }
-
-        return $data;
+        
+        // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
+        return serialize( $data );
```
tillkruss, line 2014: Always serialize, except if it already was serialized? @naxvog thoughts?

tillkruss: Does Redis currently contain double serialized data? If so, what were the use cases?

tillkruss:
```php
<?php
class Naxvog_Test {
        private $creation_time;
        private $serialization_time;
        public $some_prop;
        private $some_secret_prop;
        public function __construct() {
                $this->creation_time = time();
                $this->some_secret_prop = 'baz';
                $this->some_prop = 'bar';
        }
        public function __serialize() {
                return [
                        'serialization_time' => time(),
                        'creation_time' => $this->creation_time,
                ];
        }
        public function __unserialize( $data ) {
                $this->serialization_time = $data['serialization_time'];
                $this->creation_time = $data['creation_time'];
        }
}
add_action( 'init', function() {
        $a = new Naxvog_Test();
        $s = serialize( $a );
        #var_dump( $a, $s );
        #var_dump( unserialize( $s ) );
        if ( isset( $_GET['debug-set'] ) ) {
                wp_cache_set( 'naxvog_test_obj1', $a );
                wp_cache_set( 'naxvog_test_obj2', $s );
        }
        if ( isset( $_GET['debug-get'] ) ) {
                var_dump( wp_cache_get( 'naxvog_test_obj1' ) );
                var_dump( wp_cache_get( 'naxvog_test_obj2' ) );
        }
} );
```

tillkruss: Should we check `is_serialized()` twice in `maybe_unserialize()`?

naxvog: I would rather prefer a conversion script running on plugin update resolving such issues but this might be a non trivial approach, a LUA script and will take some time (best suited for action scheduler).

naxvog: I can't think of any valid way to serialize twice. Sure there might be serialized data within a serialized object but this is expected behaviour.

Have to look again but I'm fairly sure that I have not found any double serialized key our docker dev environment. Will have a look on my production redis instance. Should be fairly easy to find.

naxvog: OK found some instances (ran vim searching for `[[:cntrl:]]s:[[:digit:]]\+:"O` on a copy of the rdb database):

- [Plugin Update Checker](https://github.com/YahnisElsts/plugin-update-checker) (option keys `puc_external_updates-[slug]`, `puc_external_updates_theme-[slug])
- WPML (option keys: `wpml_config_index` + others)
- [Easy Social Share Buttons](https://socialsharingplugin.com/) (seems to be using Plugin Update Checker as the option key `external_updates-easy-social-share-buttons` suggests)
- [Google Sitemap Generator](https://wordpress.org/plugins/google-sitemap-generator/) (option key `sm_status`)

tillkruss: Can we simply solve this by running this twice?

https://github.com/rhubarbgroup/redis-cache/blob/fbfc86fc22b4df0c636b05e5ce5da0b548fd5b08/includes/object-cache.php#L1979-L1985

naxvog: Have to look in the code of those plugins to confirm that this is not intended. If not we should find the cause for this double serialization instead.

tillkruss:
@naxvog Thoughts on the latest push?

naxvog (changes requested):

includes/object-cache.php
tillkruss: Running `unserialize()` twice seems risky, in case plugins do their own serialization.

tillkruss:
Alright:

- We're now flush the cache right after upgrading to 2.0.16
- All data going into the object cache is serialized (strings, booleans, etc.)
- All data going out of the cache is unserialized once

includes/class-plugin.php
```diff
@@ -80,13 +80,26 @@ private function __construct() {
...
             wp_schedule_event( time(), 'hourly', 'rediscache_discard_metrics' );
         }
     }
 
+    /**
+     * Run migrations
+     *
+     * @return void
+     */
+    public function run_migrations() {
```
naxvog, line 96: To simplify things we could just flush the cache every time the plugin is updated.

Alternatively we should avoid polluting the options table with version specific entries - a general `roc_version` storing the last known version would be better. Migrations would only run if the current plugin version is newer.

naxvog (changes requested):

includes/class-plugin.php
tillkruss: Agreed 👌🏻

PR #279, opened 2020-09-21
Fixed textdomain loading too early

### LINKED ISSUE #278: Dashboard widget appears in site language instead of user language
Given a site with multiple languages installed, the "Redis Object Cache" dashboard widget always shows in the language of the site, not the chosen language of the user.

To reproduce:

* Install a non-English language that is supported by Redis Cache, for example Russian
* Change the site language to this language and ensure your translations are up to date
* Change your user language back to English from your user profile screen
* View the dashboard and observe that the widget is in Russian (the site language) instead of English (the user language)

naxvog:
Fixes #278

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -69,7 +69,6 @@ class Plugin {
     private function __construct() {
         require_once ABSPATH . 'wp-admin/includes/plugin.php';
 
-        load_plugin_textdomain( 'redis-cache', false, 'redis-cache/languages' );
         register_activation_hook( WP_REDIS_FILE, 'wp_cache_flush' );
 
         if ( is_multisite() ) {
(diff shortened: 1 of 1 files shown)


tillkruss:
Should we move the `wp_schedule_event()` as well into an `init()` method?

naxvog:
Don't know if this is strictly necessary. Keep in mind that we are loading everything at `plugins_loaded`.

tillkruss:
Not necessary, but we might as well put everything init related into a method for future use 👍🏻

naxvog:
Do you have any information regarding the `wp_schedule_event` function and that it should not be called before the `init` action? Don't see the point to delay that function to the `init` action.

tillkruss:
It's just preference.

PR #280, opened 2020-09-23
Customizations for Spinup WP

tillkruss:
Customizations for SpinupWP.com to easily rollout the full plugin, instead of a `object-cache.php` fork.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -152,13 +152,12 @@ function wp_cache_incr( $key, $offset = 1, $group = '' ) {
 function wp_cache_init() {
     global $wp_object_cache;
 
-    // Customizations: Spinup WP
-    if ( defined( 'SPINUPWP_CACHE_KEY_SALT' ) && ! defined( 'WP_REDIS_PREFIX' ) ) {
-        define( 'WP_REDIS_PREFIX', SPINUPWP_CACHE_KEY_SALT );
+    if ( ! defined( 'WP_REDIS_PREFIX' ) && getenv( 'WP_REDIS_PREFIX' ) ) {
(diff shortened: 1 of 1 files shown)


tillkruss:
@gilbitron @A5hleyRich: I can't include [this change](https://github.com/deliciousbrains/spinupwp-plugin/commit/7cc2b8c68cb3ff886431bdfe8511657f6503ac67), are you able to set `WP_REDIS_SELECTIVE_FLUSH` for deployments?

naxvog:
I don't really see the point of supporting constants with essentially different prefixes. Supporting environment variables would be a nice addition though.

gilbitron:
@tillkruss We're inclined to agree with @naxvog. If constants could be set using `getenv` that's all we would require. No need for SpinupWP prefixes. In our case, we'd need `getenv` support for `WP_REDIS_PREFIX` and `WP_REDIS_SELECTIVE_FLUSH`.

tillkruss:
@gilbitron: Done👍🏻

Anything else before we push this out?

naxvog (approved):

naxvog:
We should also allow all connection parameter and configuration constants to be set as environment variables.

(`getenv` using a parameter might be the wrong function to do this as it will return `false` if the specific variable is not present - an issue when dealing with boolean variables/constants)

tillkruss:
We can add individual ones upon request, in the meantime I'd recommend this:

```php
define( 'WP_REDIS_HOST', getenv('WP_REDIS_HOST') ?: '127.0.0.1' );
define( 'WP_REDIS_PORT', getenv('WP_REDIS_TOKEN') ?: 6379 );
```

A5hleyRich:
Awesome, thanks @tillkruss. We'll get the modifations made in SpinupWP to start distrubuting the full plugin 💪

PR #281, opened 2020-09-27
Dedicated metrics class

### LINKED ISSUE #223: Dedicated metrics class
> The [newest reported error](https://wordpress.org/support/topic/fatal-error-with-2-0-2-2/) is caused [by this line](https://github.com/rhubarbgroup/redis-cache/blob/1fbb3b48305c1dd02a7a662fb84c685e09722f9c/includes/class-plugin.php#L260).
> 
> We need to check the Redis connection status before gathering metrics and gracefully handle exceptions. Will write a dedicated class.

naxvog:
Will close #223 

Will save a serialized object of the new metrics class instead of the non standardized string to redis. Ignores entries saved using prior versions.

PS: Precursor to REST API (work in progress)

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -318,6 +318,10 @@ class Plugin {
      * @return void
      */
     public function enqueue_redis_metrics() {
+        if ( ! Metrics::is_active() ) {
+            return;
+        }
+
         $screen = get_current_screen();
 
         if ( ! isset( $screen->id ) ) {
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -8,7 +8,6 @@
 namespace Rhubarb\RedisCache;
 
 use WP_Error;
-use Exception;
 
 defined( '\\ABSPATH' ) || exit;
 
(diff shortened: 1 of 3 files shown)


tillkruss:
This is excellent.

includes/class-plugin.php
```diff
@@ -319,12 +318,6 @@ public function enqueue_admin_scripts() {
      * @return void
      */
     public function enqueue_redis_metrics() {
-        global $wp_object_cache;
-
-        if ( defined( 'WP_REDIS_DISABLE_METRICS' ) && WP_REDIS_DISABLE_METRICS ) {
-            return;
-        }
```
tillkruss, line 326: Removing this mean we're always loading `apexcharts.min.js` on the dashboard, even when metrics are disabled.

Can we call `Metrics::is_active()` here?

tillkruss (changes requested):

includes/class-plugin.php
naxvog: Good catch!
If #263 would be merged we could move the whole method into the metrics class.

tillkruss (commented):
Looks great.

PR #282, opened 2020-09-29
Reintroduction of cache key map

naxvog:
This time with considerations for multisite installations

Reference #273


tillkruss:
I'm not comfortable merging this. Too many corner cases without having tests.

PR #284, opened 2020-10-05
Adding a filter to Rhubarb\RedisCache\Plugin::validate_object_cache_dropin

rvencu:
Filter added to Rhubarb\RedisCache\Plugin::validate_object_cache_dropin function

Usage in self coded wp-content/object-cache.php file such as:

`add_filter('redis_cache_validate_object_cache_dropin', function () {
	return true;
}, 10, 2);
require_once 'plugins/redis-cache/includes/object-cache.php';`

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -402,7 +402,7 @@ class Plugin {
         $dropin = get_plugin_data( WP_CONTENT_DIR . '/object-cache.php' );
         $plugin = get_plugin_data( WP_REDIS_PLUGIN_PATH . '/includes/object-cache.php' );
 
-        return $dropin['PluginURI'] === $plugin['PluginURI'];
+        return apply_filters('redis_cache_validate_object_cache_dropin',$dropin['PluginURI'] === $plugin['PluginURI'],$dropin['PluginURI'], $plugin['PluginURI']);
     }
 
     /**


rvencu:
I also passed the two object cache files just in case someone needs to compare partial content inside them then decide to return true or false.

tillkruss:
Thanks for the PR! May I ask what changes you're making to the drop-in? Anything we should merge?

rvencu:
It is nothing too fancy, just control object cache selection for dev/stage/prod environments. Like this:

https://gist.github.com/zacscott/5e10c1e97a8fb6928bdcf8af1457fefd

tillkruss:
@naxvog: Good with you?

naxvog (changes requested):
Filter seems fine.
Please add a respective filter comment

Example:
https://github.com/rhubarbgroup/redis-cache/blob/701fdafc566e87a8b7415e216f751694b93b14ec/includes/object-cache.php#L980-L989

rvencu:
Hi, adding this comment should be my task? Please clarify...

naxvog:
If you like you can add the filter comment (without the `since` attribute as this is for @tillkruss to add).
Generally Till and you can make changes to the PR so the task is for one of you - whoever is first would be my guess :wink:.

rvencu:
Thanks. Next time I'll be careful to add proper comments but now I would prefer @tillkruss to complete it because I am not very familiar with the PR workflow and I prefer not to undo things (formatting) or mess them up. 👍

naxvog:
No need to apologize every PR is welcome. Thanks for your contribution.

tillkruss:
@rvencu Yeah, no worries. Just add a WP filter comment like the one that naxvog posted above to explain the filter to everyone else reading the code 👍🏻

rvencu:
I did something along pushing again and getting rejected, trying to rebase and ended up with my local file missing. I cannot continue so here is the comment I prepared I hope this is not a big effort for you to copy/paste into the file.

/** 
             * Filters drop-in file validation. Use with custom object cache drop-in 
            * 
            * Example usage: add these lines in custom object-cache.php file when you want to use redis-cache
            * otherwise redis-cache will be disabled and normal behaviour takes place
            * --- start sample code below ---
            * add_filter('redis_cache_validate_object_cache_dropin', function ($dropin, $original) {
            *			return true; //or compare the two arguments in some way and return bool
            *		}, 10, 2);
            *	require_once 'plugins/redis-cache/includes/object-cache.php';
            * --- end sample code above ---
            *
            * @since x.x.x 
            * @param string $dropin     the current drop-in file content
            * @param string $original   the original drop-in file content from redis-cache distribution
            */

tillkruss:
No worries. I added it to your branch.

PR #285, opened 2020-10-05
Dedicated status class

naxvog:
* Encapsulated most of the drop-in states as well as a general enabled and connected into a dedicated class
* Multiple error messages can be returned using this approach
* Added `dropin_detected` state for cases where a drop-in is present but not necessarily readable
* Deprecated the status methods in the Plugin class using a new `deprecated_function` method. Deprecation version information set to `2.1.0`
* `Plugin::get_status()` could be marked as deprecated if we would switch to `Status::get_readable( true ) ?: __( 'OK', 'redis-cache' )` 
* Deprecation function heavily borrows from WooCommerce
* Precursor to REST API (PR once this PR is merged)

Need Feedback

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -383,46 +383,37 @@ class Plugin {
     /**
      * Checks if the `object-cache.php` drop-in exists
      *
+     * @deprecated
      * @return bool
      */
     public function object_cache_dropin_exists() {
-        return file_exists( WP_CONTENT_DIR . '/object-cache.php' );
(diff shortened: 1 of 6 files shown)


tillkruss:
Neat. Should this maybe be a `Dropin` class instead?

naxvog:
Most of the state checks are for the drop-in you are right. We might be able to bundle more tests like filesystem write/copy or even certain redis states.
My intention was to give not just one error but all errors in case something is not working.

PR #286, opened 2020-10-05
Add ssl parameter for predis

MartinSilberFemale:
I just tested this for predis and I think it won't work for phpredis or credis.

So there needs to a bit more of adjustment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.0.15
+
+- Reverted `build_key()` changes due to issues in multisite environments
+
 ## 2.0.14
 
 - Made Object Cache Pro card translatable
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -5,14 +5,6 @@
     $.extend( rediscache, {
         metrics: {
             computed: null,
-            names: {
-                h: 'hits',
-                m: 'misses',
-                r: 'ratio',
-                b: 'bytes',
-                t: 'time',
-                c: 'calls',
-            },
         },
         chart: null,
         chart_defaults: {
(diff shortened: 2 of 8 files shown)


naxvog:
Hi Martin,
have you tried using the [`tls` scheme](https://github.com/rhubarbgroup/redis-cache/wiki/Connection-Parameters#wp_redis_scheme-default-tcp)?

MartinSilberFemale:
Hi @naxvog,

this is what my code looks like: 
` $redis = (object)parse_url($_ENV['REDIS_URL']);
    define('WP_REDIS_HOST', $redis->host);
    define('WP_REDIS_PORT', $redis->port);
    define('WP_REDIS_PASSWORD', $redis->pass);
    define('WP_REDIS_SCHEME', 'tls');
    define('WP_REDIS_CLIENT', 'predis');
    define('WP_REDIS_SSL', ['verify_peer' => false, 'verify_peer_name' => false]);`

I can establish the connection, when I add the changes to make the ssl call with the paramters.

Is there already another way to pass them?

naxvog:
Thanks for clarifying. So far only predis allows for SSL connection parameters to be passed. Currently there is no option to do so.

Credis and phpredis do not support those parameters so a new constant will have to be named `WP_REDIS_PREDIS_SSL` (or similar). For the same reason it might be better to check and add the connection parameter in the `connect_using_predis` method.

@tillkruss What is your take on this?

Reference: [Predis connection parameters](https://github.com/predis/predis/blob/f07f8d01fe0093e3fd9fc2a4573d119fcb8220a9/README.md#connecting-to-redis)

tillkruss:
PhpRedis supports the `tls://` scheme as of v5.3.0 and it also supports custom TLS options with a similar syntax passed into the constructor.

Credis seems to support TLS as well, but I haven't looked into it.

naxvog:
Have not found any documentation on PhpRedis' part ([only found the `tls://` scheme](https://github.com/phpredis/phpredis#pconnect-popen)) regarding SSL parameters but I have looked in the sourcecode of Credis and as far as I saw there are none.

tillkruss:
Both describe the scheme in their readme:

https://github.com/phpredis/phpredis#connect-open

https://github.com/colinmollenhour/credis#tls-connection-string

I know PhpRedis supports socket options like Predis does. Not sure with Credis.

tillkruss:
@MartinSilberFemale: Are you able to add support for PhpRedis as well in this PR?

MartinSilberFemale:
@tillkruss Yeah if think I could try that, but would take some time, if this is okay for you :) (not more than 7 days)

tillkruss:
Any update on this?

MartinSilberFemale:
@tillkruss Sorry, that I didn't reply. 

Right now I am really busy with other projects, but I will try to take a look at it next week.

PR #296, opened 2020-11-05
isset() does not allow null

szepeviktor:
Please see https://www.php.net/manual/en/function.isset.php

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -514,7 +514,7 @@ class WP_Object_Cache {
         if ( isset( $parameters['password'] ) ) {
             $password = $parameters['password'];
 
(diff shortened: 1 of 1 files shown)


tillkruss:
Why do you want to remove the check? The password might be `null`.

szepeviktor:
> The password might be `null`.

@tillkruss The password is `null` `if ( isset( $parameters['password'] ) ) {` will not run
as _isset() does not allow null_

tillkruss:
Gotcha, the code made no sense. @naxvog approved?

szepeviktor:
@phpstan taught me that `isset` work this way.

szepeviktor:
# ✔️

PR #300, opened 2020-11-20
Do not display the dropin message if redis is disabled.

leopoiroux:
I would like to control when I can replace object-cache.php. As long as WP_REDIS_DISABLED, I want to be sure that no user of my site can overwrite object-cache.php (we are several admin)


tillkruss:
Can you allow maintainer edits, so I can fix the formatting?

leopoiroux:
I moved my PR. Maintainer edits seems not to be allowed if the fork is made within an organisation.

PR #302, opened 2020-11-23
Do not display the dropin message if WP_REDIS_DISABLE_DROPIN_BANNERS …

leopoiroux:
…exists and true.

I would like to control when I can replace object-cache.php. As long as WP_REDIS_DISABLE_DROPIN_BANNERS, I want to be sure that no user of my site can overwrite object-cache.php (we are several admin)

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -571,6 +571,11 @@ class Plugin {
         if ( ! current_user_can( is_multisite() ? 'manage_network_options' : 'manage_options' ) ) {
             return;
         }
+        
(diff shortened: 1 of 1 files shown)


leopoiroux:
I moved my PR. Maintainer edits seems not to be allowed if the fork is made within an organisation.

tillkruss:
@leopoiroux: Thanks!

tillkruss:
I'll tag a new version some time this week.

leopoiroux:
👍

PR #303, opened 2020-11-25
Added constant WP_REDIS_DISABLE_DROPIN_MANAGE

leopoiroux:
I had previously done a PR to hide the dropin banner. This code was not complete. I need a wider deactivation of the dropin management. In my case, I want to master the dropin code to install and when.
So I replaced my constant by WP_REDIS_DISABLE_DROPIN_MANAGE

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -576,7 +576,6 @@ class Plugin {
             return;
         }
         
-        // Do not display the dropin message if you want
         if ( defined( 'WP_REDIS_DISABLE_DROPIN_MANAGE' ) && WP_REDIS_DISABLE_DROPIN_MANAGE ) {
             return;
         }
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -398,6 +398,10 @@ class Plugin {
         if ( ! $this->object_cache_dropin_exists() ) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


leopoiroux:
Why I need to manage the "object-cache.php" file: today our Wordpress is shared between 100 sites. We would like to migrate to redis instead of Memcached. However, we want to control this changeover. Our object-cache.php file adapts itself according to the site to point to Redis or Memcached. 

Without the WP_REDIS_DISABLE_DROPIN_MANAGE option, our object-cache.php file is not recognised and can be deleted or overwritten when activating/deactivating the plugin.

tillkruss:
Sorry, the issue is that the drop-in is deleted when the Redis Object Cache plugin is disabled?

leopoiroux:
Yes the drop-in is deleted when the plugin is deactivated and is overwritten when the plugin is activated. My constant is used to indicate that you want to keep control over this file.

tillkruss:
Isn't `WP_REDIS_DISABLE_DROPIN_AUTOUPDATE` enough?

https://github.com/rhubarbgroup/redis-cache/blob/3f7dd66cddd0c4316a4ccea0f6ae6c00196a5088/includes/class-plugin.php#L998-L1000

PR #316, opened 2021-03-12
add 'redis_object_cache_trace' hook. Enables tracing of every call, useful for fine grained metric collection

fleandro-obs:
The proposal is simple, add a new hook that reports statistics (hits & misses) for every cache call.

This enables end users to implement fine grained metrics in order to better understand cache runtime behaviour, in order to optimize cache usage.

The proposed signature is as follows:
```php
/**
    * Fires on every cache call.
    *
    * This hook is called on every cache request.
    * It reports statistics per key involved. @see WP_Object_Cache::TRACE_FLAG_READ and friends.
    *
    * @param  string             $command   The command that was executed.
    * @param  string             $group     Key group.
    * @param  array[string]array $keyValues Maps keys to the returned values (if any) and the resulting status.
    *  $keyValues = [
    *      "foo" => ["value" => "bar", "status" => TRACE_FLAG_READ | TRACE_FLAG_HIT],                       // hit on redis (implies internal miss)
    *      "baz" => ["value" => "quo", "status" => TRACE_FLAG_READ | TRACE_FLAG_HIT | TRACE_FLAG_INTERNAL], // hit on internal cache
    *      "eta" => ["value" => null,  "status" => TRACE_FLAG_READ],                                        // miss
    * ];
    * @param  float              $duration  Duration of the request in microseconds.
    * @return void
*/
do_action( 'redis_object_cache_trace', $command, $group, $keyValues, $duration );
```

This signature is able to accommodate for both batch and single key flows.
For every key, we pass the value as well as the operation metadata to the user. The user is free to decide which metadata is relevant for their use case.

Flags are defined as follows:
```php
/**
    * Operation pertains to internal cache, not Redis.
    *
    * @since x.x.x
    * @var int
*/
const TRACE_FLAG_INTERNAL = 1 << 0;
/**
    * Operation resulted in a cache hit.
    *
    * @since x.x.x
    * @var int
*/
const TRACE_FLAG_HIT = 1 << 1;
/**
    * Read operation.
    *
    * @since x.x.x
    * @var int
*/
const TRACE_FLAG_READ = 1 << 2;
/**
    * Write operation.
    *
    * @since x.x.x
    * @var int
*/
const TRACE_FLAG_WRITE = 1 << 3;
/**
    * Delete operation.
    *
    * @since x.x.x
    * @var int
*/
const TRACE_FLAG_DEL = 1 << 4;
/**
    * Operation bypassed internal cache.
    *
    * @since x.x.x
    * @var int
*/
const TRACE_FLAG_REFRESH = 1 << 5;
```

Intended usage is as follows:
```php
add_action( 'redis_object_cache_trace', function($command, $group, $keyValues, $duration) {
    foreach ($keyValues as $key => $result) {
        $flags = $result['status'];
        $value = $result['value'];
       
        $isInternal = ($flags & WP_Object_Cache::TRACE_FLAG_INTERNAL) > 0;
        $isHit = ($flags & WP_Object_Cache::TRACE_FLAG_HIT) > 0;
        $isReadOp = ($flags & WP_Object_Cache::TRACE_FLAG_READ) > 0;
        $isWriteOp = ($flags & WP_Object_Cache::TRACE_FLAG_WRITE) > 0;
        $isDelOp = ($flags & WP_Object_Cache::TRACE_FLAG_DEL) > 0;
        $isRefresh = ($flags & WP_Object_Cache::TRACE_FLAG_REFRESH) > 0;

        
        // do stuff.
    }
}, 10, 4);
```

We decided to ignore writes to the internal cache, as we don't see any value in measuring it. If you feel like we should, even if it's for the sake of consistency alone, let us know.

I'm not a PHP developer, nor am I that familiar with WP, so please let me know if there's anything that can be improved.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1068,7 +1068,7 @@ class WP_Object_Cache {
                         'value' => $value,
                         'status' => self::TRACE_FLAG_WRITE,
                     ],
-                ], microtime(true) - $start_time );
+                ], microtime( true ) - $start_time );
 
                 $this->cache_calls++;
                 $this->cache_time += $execute_time;
@@ -1408,7 +1408,7 @@ LUA;
                     'value' => $value,
                     'status' => $trace_flags | self::TRACE_FLAG_HIT | self::TRACE_FLAG_INTERNAL,
                 ],
-            ], microtime(true) - $start_time);
+            ], microtime( true ) - $start_time);
 
             return $value;
         } elseif ( $this->is_ignored_group( $group ) || ! $this->redis_status() ) {
@@ -1420,7 +1420,7 @@ LUA;
                     'value' => null,
                     'status' => $trace_flags | self::TRACE_FLAG_INTERNAL,
                 ],
-            ], microtime(true) - $start_time );
+            ], microtime( true ) - $start_time );
 
             return false;
         }
@@ -1448,7 +1448,7 @@ LUA;
                     'value' => null,
                     'status' => $trace_flags,
                 ],
-            ], microtime(true) - $start_time );
+            ], microtime( true ) - $start_time );
             return false;
         } else {
             $found = true;
@@ -1463,7 +1463,7 @@ LUA;
                 'value' => $value,
                 'status' => $trace_flags | self::TRACE_FLAG_HIT,
             ],
-        ], microtime(true) - $start_time );
+        ], microtime( true ) - $start_time );
 
         if ( function_exists( 'do_action' ) ) {
             /**
@@ -1521,7 +1521,7 @@ LUA;
 
         $cache = [];
         $derived_keys = [];
-        $start_time = microtime(true);
+        $start_time = microtime( true );
 
         foreach ( $keys as $key ) {
             $derived_keys[ $key ] = $this->build_key( $key, $group );
@@ -1557,7 +1557,7 @@ LUA;
                 }
             }
 
-            $this->trace_command( 'mget', $group, $traceKV, microtime(true) - $start_time );
+            $this->trace_command( 'mget', $group, $traceKV, microtime( true ) - $start_time );
 
             return $cache;
         }
@@ -1599,7 +1599,7 @@ LUA;
         );
 
         if ( empty( $remaining_keys ) ) {
-            $this->trace_command( 'mget', $group, $traceKV, microtime(true) - $start_time );
+            $this->trace_command( 'mget', $group, $traceKV, microtime( true ) - $start_time );
 
             return $cache;
         }
@@ -1796,7 +1796,7 @@ LUA;
                     'value' => $value,
                     'status' => $trace_flags | self::TRACE_FLAG_INTERNAL,
                 ],
-            ], microtime(true) - $start_time );
+            ], microtime( true ) - $start_time );
 
             return $value;
         }
@@ -1864,7 +1864,7 @@ LUA;
                     'value' => $value,
                     'status' => $trace_flags | self::TRACE_FLAG_INTERNAL,
                 ],
-            ], microtime(true) - $start_time );
+            ], microtime( true ) - $start_time );
 
             return $value;
         }
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -346,6 +346,13 @@ class WP_Object_Cache {
      *
      * @var array
      */
+    public $trace_enabled = false;
+
+    /**
+     * Holds the error messages.
+     *
+     * @var array
+     */
     public $errors = [];
 
     /**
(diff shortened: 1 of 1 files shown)


includes/object-cache.php
```diff
@@ -2138,6 +2242,50 @@ protected function handle_exception( $exception ) {
...
     }
+
+    /**
+     * Calculates how much time has elapsed since $start_time.
+     *
+     * @param float $start_time
+     * @return float
+     */
+    private function elapsed( $start_time ) {
+        return microtime(true) - $start_time;
```
tillkruss, line 2253: If `microtime()` isn't good enough, we could use `hrtime(true) * 1e-9`.

includes/object-cache.php
```diff
@@ -2138,6 +2242,50 @@ protected function handle_exception( $exception ) {
...
+     * @param  string             $command
+     * @param  string             $group
+     * @param  array[string]array $keyValues
+     * @param  float              $duration
+     * @return void
+     */
+    private function trace_command ( $command, $group, $keyValues, $duration ) {
+        if ( !function_exists( 'do_action' ) ) {
+            return;
+        }
```
tillkruss, line 2268: In addition to a `do_action()` check, we should probably also call `has_action('redis_object_cache_trace')`.

https://developer.wordpress.org/reference/functions/has_action/

tillkruss (changes requested):

tillkruss:
@fleandro-obs This is excellent work.

@naxvog Any thoughts?

includes/object-cache.php
naxvog: Sidenote: checking `has_action` will avoid firing the `all` hook.  Performance would be gained by checking this first but it might be unexpected.

includes/object-cache.php
naxvog: `hrtime` is only available since PHP 7.3 so we should not use it for now but it might be a good feature addition candidate for version 3

naxvog:
I for one would love to see this functionality implemented in a standalone static class only being loaded if a configuration constant is present in order to avoid bloating the main class further.

tillkruss:
I thought so too. Quick constant check before calling any extra methods.

fleandro-obs:
sorry for the radio silence, on vacation at the moment, planning to address your feedback next week

tillkruss:
@fleandro-obs: No rush, next week is great 👍

fleandro-obs:
> I for one would love to see this functionality implemented in a standalone static class only being loaded if a configuration constant is present in order to avoid bloating the main class further.

@naxvog what kind of bloat are we trying to address, declaration/api bloat? If that's the case how would you feel about extracting the flags into a separate "enum" and leaving the rest as is?

I fear that anything more than that might prove counterproductive. This does not introduce any new functionality, it is merely an introspection capability. As such, it's inherently intertwined with the implementation. Creating a layer of separation between the two would require us to add a new branch on every call site, which might feel much more "bloaty", both in terms of the code itself but also performance wise.

Aside: I suspect that most PHP implementations would benefit very little from branch prediction amortizing the cost, which makes me a little hesitant to create more branches. This is just a hunch tho, I have 0 evidence to support this, feel free to set me straight :smile:

tillkruss:
@fleandro-obs: I think he means to bloat of WordPress API calls. As long as we address this, we should be good:
https://github.com/rhubarbgroup/redis-cache/pull/316/files#r593347999

fleandro-obs:
> @fleandro-obs: I think he means to bloat of WordPress API calls. As long as we address this, we should be good:
> https://github.com/rhubarbgroup/redis-cache/pull/316/files#r593347999

Ah! Roger that, I'll fix it next moday.

fleandro-obs:
@tillkruss updated the PR to address your feedback, let me know if you need anything else

tillkruss:
@fleandro-obs: I made some smaller modifications. Before I merge this I'd like to get rid of `trace_enabled()`, maybe just be using a `WP_REDIS_TRACE` constant.

Can you tell me more about _where_ and _when_ you're registering the `redis_object_cache_trace` action?

fleandro-obs:
> @fleandro-obs: I made some smaller modifications. Before I merge this I'd like to get rid of `trace_enabled()`, maybe just be using a `WP_REDIS_TRACE` constant.

Do you mean dropping the `has_action( 'redis_object_cache_trace' )` and use `defined(WP_REDIS_TRACE)` in conjunction with `function_exists( 'do_action' )`?

> 
> Can you tell me more about _where_ and _when_ you're registering the `redis_object_cache_trace` action?

I'll try, please let me know if I get the lingo wrong. Currently we're using it in the theme's `functions.php`. The plan is to move it to an mu-plugin once we can clean up our composer setup (it's what's stopping us currently).
As to how, we're filtering this information and exposing it as prometheus metrics.

tillkruss:
Yeah, a must-use plugin would certainly be better, the theme is loaded way later and you'll miss out on some of the traces that plugins and WP core does.

Shameless plug: [Object Cache Pro](https://objectcache.pro) allows you to log/trace all commands out of the box 🤭

Using a constant like `WP_REDIS_TRACE` would be more efficient, I'll add that.

tillkruss (approved):

tillkruss:
@fleandro-obs Would you mind testing `develop` in your production/staging site to see if it works as expected?

fleandro-obs:
hi @tillkruss,

when you removed the $trace_enabled variable in `get_multiple` you forgot to replace it by `$this->trace_enabled` in a few places, for eg: https://github.com/fleandro-obs/redis-cache/blob/ec76171aff4b751121bd9bf98920ac0fcfc33ef9/includes/object-cache.php#L1596

Other than that no issues so far :+1:

tillkruss:
@fleandro-obs: Nice catch, glad this is working well 👌

PR #345, opened 2022-04-04
Update add_non_persistent_groups()

### LINKED ISSUE #74: WP_REDIS_IGNORED_GROUPS: Why is `counts` in there? I want to cache that group
`WP_REDIS_IGNORED_GROUPS ` (formerly `->no_redis_groups`) controls the part of the object cache related to the `wp_cache_add_non_persistent_groups` function in core. 

Groups added to this constant and thus registered in the plugin don't get cached ever. 

On our site, this ended up being the biggest slowdown, because `wp_count_posts` uses that group to cache it's query, and for us that query is really slow because we have sooo many posts. Because that query runs on basically every page on the site, it would be enormously valuable to have it in the cache. 

**So the question is this: Why is it in `WP_REDIS_IGNORED_GROUPS` and can it be removed?**

FWIW based on our research I suspect this is a question with a perfectly good answer, but I wanted to create this ticket anyway so there's at least a reference. I'll update this original post with as much info as I can find. 

---

### History of inclusion in this plugin
* In terms of this plugin, the inclusion of `counts` in the ignored list goes all the way back to [when you imported ehitter's version](https://github.com/tillkruss/redis-cache/commit/dc5cb26202cb22a609149d9b5e0d34c0a6d92e75#diff-3ea3a8942f84c506446c32d498ac655d). 
* ehitter's version of the plugin had the hardcoded `no_redis_groups` array all the way back to when [it was imported from a Memcached plugin](https://github.com/ericmann/Redis-Object-Cache/commit/baa6de5240e41d26f49c6c5ee86da24583c49e83)

So it goes all the way back and was never something this plugin made as a conscious decision. Good to know. 

In any event, digging deeper into the core code we find that the core pluggable function `wp_start_object_cache()` (which is overridden by this plugin's function) is using the same defaults anyway: 

```
	wp_cache_add_non_persistent_groups( array( 'counts', 'plugins' ) );
```

And, basically, the object caching plugins are just re-implementing this blacklist from core. 

Okay so to me this implies that most likely it just isn't appropriate to have the `counts` group ever get persistently cached, and that core never really promised they would be. 

Specifically, the logic here seems to be that they want the counts cached WITHIN each page, but to have them re-generated on each new pageload. 

The problem for me still is I don't know why `counts` can't be persistently cached. `wp_start_object_cache()` has no comments anywhere near the call to `wp_cache_add_non_persistent_groups` so it's not at all clear why this hardcoded list is present.

### LINKED ISSUE #114: Caching "counts" cache group
Hi, 

I am trying to remove `counts` from ignored cached groups. I did add
`define( 'WP_REDIS_IGNORED_GROUPS', ['plugins'] );` and checked if the constant is being loaded properly and it works. However it turns out that Wordpress runs `wp_cache_add_non_persistent_groups( array( 'counts', 'plugins' ) );` in wp-includes/load.php on line 586 (https://core.trac.wordpress.org/browser/branches/5.2/src/wp-includes/load.php#L586) which adds `counts` to ignored groups back regardless of the `WP_REDIS_IGNORED_GROUPS` setting.

What should I do to enable `counts` caching? Also, do you have any idea why counts caching is disabled by default? I did study the `wp_count_posts()` function (https://core.trac.wordpress.org/browser/branches/5.2/src/wp-includes/post.php#L2527) that sets this cache key and it seems to handle all the invalidation etc. so it seems really wasteful to recount posts on every request by default. This was already discussed in #74 but no solution was provided there.

### LINKED ISSUE #346: counts group cannot be removed from ignored groups
For details see: 

*  #114
* #74
* https://core.trac.wordpress.org/ticket/47884#ticket

I've created a PR that addresses the issue in this plugin as was advised by WP Core team in the above linked trac ticket.

lots0logs:
Don't allow groups that were manually removed using `WP_REDIS_IGNORED_GROUPS` constant to be added to the ignored groups list.

* Fixes #346 
* FIxes #114
* Fixes #74
* Ref: https://core.trac.wordpress.org/ticket/47884#ticket


includes/object-cache.php
```diff
@@ -2417,7 +2417,21 @@ public function add_global_groups( $groups ) {
...
     public function add_non_persistent_groups( $groups ) {
-        $groups = (array) $groups;
+        $groups          = (array) $groups;
+        $default_groups  = [ 'counts', 'plugins', 'themes' ];
+
+        static $groups_override        = null;
+        static $removed_default_groups = null;
+
+        if ( is_null( $groups_override ) && defined( 'WP_REDIS_IGNORED_GROUPS' ) && is_array( WP_REDIS_IGNORED_GROUPS ) ) {
+            $groups_override        = array_map( [ $this, 'sanitize_key_part' ], WP_REDIS_IGNORED_GROUPS );
```
tillkruss, line 2427: `sanitize_key_part()` is no longer needed here.

includes/object-cache.php
```diff
@@ -2417,7 +2417,21 @@ public function add_global_groups( $groups ) {
      * @param array $groups  List of groups that are to be ignored.
      */
     public function add_non_persistent_groups( $groups ) {
-        $groups = (array) $groups;
+        $groups          = (array) $groups;
+        $default_groups  = [ 'counts', 'plugins', 'themes' ];
```
tillkruss, line 2421: Not stoked about keeping track here of these. I rather ditch `$this->ignored_groups` and get everything from core.

tillkruss (changes requested):

includes/object-cache.php
lots0logs: I agree it's not ideal but it's the only way I can think of handling this that doesn't break backwards compatibility.

includes/object-cache.php
lots0logs: I copied that from what was done in `__construct()` for `$this->ignored_groups`. Sure you want to remove it? Let me know and I'll make the change.

tillkruss: Yes, I removed it in the 2.1.x release.

includes/object-cache.php
tillkruss: Okay, let's at least use the class property.

includes/object-cache.php
lots0logs: I synced my branch with `develop` and still see it being done in `__construct()` :thinking: :

https://github.com/lots0logs/redis-cache/blob/patch-1/includes/object-cache.php#L490-L492

includes/object-cache.php
lots0logs: Done :heavy_check_mark:

lots0logs:
I decided to ditch my previous implementation and just add a filter instead. Much easier this way, IMO.

PR #349, opened 2022-05-27
fix: prevent errors on diagnostics page if the plugin is disabled

rkaiser0324:
If the plugin is disabled via `WP_REDIS_DISABLED`, then errors like this are thrown on the diagnostic page:

`An error of type E_ERROR was caused in line 127 of the file /path/to/redis-cache/includes/diagnostics.php`

`Error message: Uncaught TypeError: array_values(): Argument #1 ($array) must be of type array, null`

This prevents them.


tillkruss:
Thanks Rolf.

PR #352, opened 2022-06-29
Enhance `add_multiple` & `set_multiple` functionality

zaherg:
This PR rewrite the `add_multiple` and `set_multiple`

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -2287,28 +2287,11 @@ LUA;
         }
 
         foreach ($data as $key => $value) {
-            $start_time = microtime( true );
             $key = $this->sanitize_key_part( $key );
             $derived_key = $this->fast_build_key( $key, $group );
 
             if ( isset( $values[ $key ] ) && $values[ $key ] ) {
                 $this->add_to_internal_cache( $derived_key, $value );
-            }
-
-            if ( function_exists( 'do_action' ) ) {
-                $execute_time = microtime( true ) - $start_time;
-
-                /**
-                 * Fires on every cache set multiple request
-                 *
-                 * @since 1.2.2
-                 * @param string $key          The cache key.
-                 * @param mixed  $value        Value of the cache entry.
-                 * @param string $group        The group value appended to the $key.
-                 * @param int    $expiration   The time in seconds the entry expires. 0 for no expiry.
-                 * @param float  $execute_time Execution time for the request in seconds.
-                 */
-                do_action( 'redis_object_cache_set_multiple', $key, $value, $group, $expiration, $execute_time );
             }
         }
 
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1284,9 +1284,7 @@ class WP_Object_Cache {
      * @return  bool                   Returns TRUE on success or FALSE on failure.
      */
     protected function add_or_replace( $add, $key, $value, $group = 'default', $expiration = 0 ) {
-        $cache_addition_suspended = function_exists( 'wp_suspend_cache_addition' )
-            ? wp_suspend_cache_addition()
-            : false;
+        $cache_addition_suspended = function_exists( 'wp_suspend_cache_addition' ) && wp_suspend_cache_addition();
 
         if ( $add && $cache_addition_suspended ) {
             return false;
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -2270,7 +2270,7 @@ LUA;
                         ];
                     }, $keys ));
 
-                    $this->trace_command( 'pipeline', $group, $traceKV, $execute_time );
+                    $this->trace_command( 'set', $group, $traceKV, $execute_time );
                 }
             }
 
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -2172,10 +2172,10 @@ LUA;
     /**
      * Sets multiple values to the cache in one call.
      *
-     * @param array  $data   Array of key and value to be set.
-     * @param string $group  Optional. Where the cache contents are grouped.
-     * @param int    $expire Optional. When to expire the cache contents, in seconds.
-     *                       Default 0 (no expiration).
+     * @param array  $data       Array of key and value to be set.
+     * @param string $group      Optional. Where the cache contents are grouped.
+     * @param int    $expiration Optional. When to expire the cache contents, in seconds.
+     *                           Default 0 (no expiration).
      * @return bool[] Array of return values, grouped by key. Each value is always true.
      */
     public function set_multiple( array $data, $group = 'default', $expiration = 0 ) {
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1190,6 +1190,8 @@ class WP_Object_Cache {
         $group = $this->sanitize_key_part( $group );
 
         if ( ! $this->is_ignored_group( $group ) ) {
+            $start_time = microtime( true );
+
             $orig_exp = $expire;
             $expire = $this->validate_expiration( $expire );
             $tx = $this->redis->pipeline();
@@ -1228,8 +1230,6 @@ class WP_Object_Cache {
 
                 $tx->set( ...$args );
             }
-
-            $start_time = microtime( true );
 
             try {
                 $method = ( $this->redis instanceof Predis\Client ) ? 'execute' : 'exec';
@@ -1485,6 +1485,9 @@ class WP_Object_Cache {
      *                true on success, or false if the contents were not deleted.
      */
     protected function delete_multiple_at_once( array $keys, $group = 'default' ) {
+        $start_time = microtime( true );
+        $traceKV = [];
+
         if ( $this->is_ignored_group( $group ) ) {
             $results = [];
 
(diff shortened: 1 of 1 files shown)


includes/object-cache.php
```diff
@@ -1160,13 +1160,77 @@ public function add_multiple( array $data, $group = 'default', $expire = 0 ) {
...
+
+                $results[$key] = false;
+            }
 
-        foreach ( $data as $key => $value ) {
-            $values[ $key ] = $this->add( $key, $value, $group, $expire );
         }
 
-        return $values;
+        return $results;
```
tillkruss, line 1242: I think we should do the same wrapper:

https://github.com/rhubarbgroup/redis-cache/blob/56ff6116eb00cf4dc54e5bff1fcbce2d238dcd65/includes/object-cache.php#L1370-L1382

includes/object-cache.php
```diff
@@ -1160,13 +1160,86 @@ public function add_multiple( array $data, $group = 'default', $expire = 0 ) {
...
+                }, $tx->{$method}() );
+
+                if ( $results ) {
+                    $results = array_combine( $keys, $results );
+                }
+
+            } catch ( Exception $exception ) {
+                $this->handle_exception( $exception );
+
+                $results[$key] = false;
```
tillkruss, line 1237: Missing spacing.

includes/object-cache.php
```diff
@@ -1160,13 +1160,86 @@ public function add_multiple( array $data, $group = 'default', $expire = 0 ) {
...
         }
 
-        $values = [];
+        $results = [];
+
+        if (! $this->is_ignored_group($group) && $this->redis_status() ) {
+            $orig_exp = $expire;
+            $expire = $this->validate_expiration( $expire );
+            $tx = $this->redis->pipeline();
+            $keys = array_keys($data);
```
tillkruss, line 1169: Missing spacing.

includes/object-cache.php
```diff
@@ -1160,13 +1160,86 @@ public function add_multiple( array $data, $group = 'default', $expire = 0 ) {
             return array_combine( array_keys( $data ), array_fill( 0, count( $data ), false ) );
         }
 
-        $values = [];
+        $results = [];
+
+        if (! $this->is_ignored_group($group) && $this->redis_status() ) {
```
tillkruss, line 1165: Missing spacing.

includes/object-cache.php
```diff
@@ -1160,13 +1160,86 @@ public function add_multiple( array $data, $group = 'default', $expire = 0 ) {
...
+                 * @since 1.4.2
+                 * @param int    $expiration The time in seconds the entry expires. 0 for no expiry.
+                 * @param string $key        The cache key.
+                 * @param string $group      The cache group.
+                 * @param mixed  $orig_exp   The original expiration value before validation.
+                 */
+                $expire = apply_filters( 'redis_cache_expiration', $expire, $key, $group, $orig_exp );
+                $start_time = microtime( true );
+
+                $derived_key = $this->fast_build_key( $key, $group );
```
tillkruss, line 1184: I think this needs the `sanitize_key_part();` calls as well.

tillkruss (changes requested):

includes/object-cache.php
zaherg: you mean using 
```
     if ( $this->redis_status() && method_exists( $this->redis, 'pipeline' ) ) { 
         return $this->delete_multiple_at_once( $keys, $group ); 
     } 
```

or using `values` instead of `results`

tillkruss: Just two methods, one with the fallback of pipeline() isn't supported.

tillkruss: Let's use the same approach with a dedicated `*_multiple_at_once()` method to make this more readable.

zaherg: @tillkruss done, can you check.

includes/object-cache.php
```diff
@@ -1169,6 +1173,111 @@ public function add_multiple( array $data, $group = 'default', $expire = 0 ) {
...
+     * @return bool[] Array of return values, grouped by key. Each value is either
+     *                true on success, or false if cache key and group already exist.
+     */
+    protected function add_multiple_at_once( array $data, $group = 'default', $expire = 0 )
+    {
+        $values = [];
+
+        $group = $this->sanitize_key_part( $group );
+
+        if (! $this->is_ignored_group( $group ) && $this->redis_status() ) {
```
tillkruss, line 1192: This is missing support for ignored groups like `delete_multiple_at_once()` has.

tillkruss: Doesn't need to check for `redis_status()` again.

includes/object-cache.php
```diff
@@ -2104,6 +2215,141 @@ public function set_multiple( array $data, $group = 'default', $expiration = 0 )
...
+    /**
+     * Sets multiple values to the cache in one call.
+     *
+     * @param array  $data       Array of key and value to be set.
+     * @param string $group      Optional. Where the cache contents are grouped.
+     * @param int    $expiration Optional. When to expire the cache contents, in seconds.
+     *                           Default 0 (no expiration).
+     * @return bool[] Array of return values, grouped by key. Each value is always true.
+     */
+    public function set_multiple_at_once( array $data, $group = 'default', $expiration = 0 )
```
tillkruss, line 2227: This should be protected and then doesn't need to check for `redis_status()` again.

includes/object-cache.php
```diff
@@ -2104,6 +2215,141 @@ public function set_multiple( array $data, $group = 'default', $expiration = 0 )
...
+    {
+        $values = [];
+
+        $group = $this->sanitize_key_part( $group );
+
+        // Save if group not excluded from redis and redis is up.
+        if (! $this->is_ignored_group( $group ) && $this->redis_status() ) {
```
tillkruss, line 2234: This is missing support for ignored groups like `delete_multiple_at_once()` has.

includes/object-cache.php
```diff
@@ -2104,6 +2215,141 @@ public function set_multiple( array $data, $group = 'default', $expiration = 0 )
...
+
+            $start_time = microtime( true );
+
+            foreach ( $data as $key => $value ) {
+                $key = $this->sanitize_key_part( $key );
+
+                $derived_key = $this->fast_build_key( $key, $group );
+
+                // Save if group not excluded from redis and redis is up.
+                if ( ! $this->is_ignored_group( $group ) && $this->redis_status() ) {
```
tillkruss, line 2250: There is only one group, not need for this redundant check.

tillkruss (changes requested):

includes/object-cache.php
zaherg: > This is missing support for ignored groups like `delete_multiple_at_once()` has.

I guess you mean we should add them to the internal cache like in the `add_or_replace` function not redis.

https://github.com/zaherg/redis-cache/blob/better-implementation-for-add-multiple-and-set-multiple/includes/object-cache.php#L1400-L1404

includes/object-cache.php
```diff
@@ -2104,6 +2203,101 @@ public function set_multiple( array $data, $group = 'default', $expiration = 0 )
...
+
+            $execute_time = microtime( true ) - $start_time;
+            $this->cache_calls++;
+            $this->cache_time += $execute_time;
+
+            if ( $this->trace_enabled ) {
+                $traceKV = array_merge( ...array_map( function ( $key ) {
+                    return [
+                        $key => [
+                            'value' => null,
```
tillkruss, line 2279: Can we get that value added?

includes/object-cache.php
```diff
@@ -2088,13 +2183,17 @@ public function set( $key, $value, $group = 'default', $expiration = 0 ) {
     /**
      * Sets multiple values to the cache in one call.
      *
-     * @param array  $data   Array of key and value to be set.
-     * @param string $group  Optional. Where the cache contents are grouped.
-     * @param int    $expire Optional. When to expire the cache contents, in seconds.
-     *                       Default 0 (no expiration).
+     * @param array  $data       Array of key and value to be set.
+     * @param string $group      Optional. Where the cache contents are grouped.
+     * @param int    $expiration Optional. When to expire the cache contents, in seconds.
```
tillkruss, line 2188: Let's change the parameter name to `$expire`. Nice catch.

includes/object-cache.php
```diff
@@ -2104,6 +2203,101 @@ public function set_multiple( array $data, $group = 'default', $expiration = 0 )
...
+                    return (bool) $this->parse_redis_response( $response );
+                }, $tx->{$method}() );
+
+                if ( $values ) {
+                    $values = array_combine( $keys, $values );
+                }
+            } catch ( Exception $exception ) {
+                $this->handle_exception( $exception );
+
+                return array_combine( $keys, array_fill( 0, count( $keys ), false ) );
```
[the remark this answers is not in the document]
zaherg: I did test it with ignored group, but not with an error tbh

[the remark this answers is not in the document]
tillkruss: No worries, just test it once for both functions 👍

PR #361, opened 2022-07-13
Remove PHP tools from CI

szepeviktor:
Composer v2 is the default in shivammathur/setup-php

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,8 +1,8 @@
 name: Run Tests
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -1,8 +1,8 @@
 name: Run Tests
 
 on:
-  push:
-  pull_request:
+  push: null
+  pull_request: null
```
szepeviktor, line 5: Default YAML value.

PR #364, opened 2022-07-18
Enhanced testing

zaherg:
This is an opinionated enhancement for the testing, instead of using docker we pull wordpress test-libs and use it for out feature testing.

The tests are still the same old one, except a few changes, so they are all accredited to the [original creator](https://github.com/rhubarbgroup/redis-cache/pull/329) but due to the changes I made, the ownership has changed.

We can use this as a base to build upon and add more tests.

> **Warning** This is an opinionated enhancement
>
> You should run `./setup-test.sh` to setup the test environment locally.
> If you want to change any of they the test environment variable, create a new file called `config.sh` and add the following to it:
> 
> ```bash
> GIT_BRANCH=trunk # the wordpress version you want to test against
> DB_HOST=127.0.0.1
> DB_PASS=mysuperpassword # your password, default empty
> DB_NAME=yourdatabase # your database name, default `wp_tests`
> DB_USER=youruser # your database user, default `root`
> ```
> 
> Creating `config.sh` file is not mandatory, as long as you are using the default values.


tests/bootstrap.php
```diff
@@ -0,0 +1,65 @@
...
+
+// function get_file_data($file, $header)
+// {
+//     return ['Version' => '2.1.0'];
+// }
+
+// function register_activation_hook(string $file, callable $callback)
+// {
+//     return true;
+// }
```
tillkruss, line 65: Can we remove these?

tests/TestCase.php
```diff
@@ -0,0 +1,38 @@
...
+        // Your own additional tear down.
+        parent::tear_down();
+    }
+
+    protected function init_cache()
+    {
+        global $wp_object_cache;
+        $cache_class = get_class($wp_object_cache);
+        $cache = new $cache_class();
+        $cache->add_global_groups(['global-cache-test', 'users', 'userlogins', 'usermeta', 'user_meta', 'useremail', 'userslugs', 'site-transient', 'site-options', 'blog-lookup', 'blog-details', 'rss', 'global-posts', 'blog-id-cache', 'networks', 'sites', 'site-details']);
```
tillkruss, line 34: Let's use new lines here.

tests/ExampleTest.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Unit;
+
+use Tests\TestCase;
+
+/**
+ * @internal
+ * @coversNothing
+ */
+class ExampleTest extends TestCase
```
tillkruss, line 13: Should we drop this file?

setup-test.sh
```diff
@@ -0,0 +1,48 @@
+#! /usr/bin/env sh
```
tillkruss, line 1: Let's put this file in `bin`.

README.md
```diff
@@ -48,4 +48,24 @@ To see a list of all available WP-CLI commands, please see the [WP CLI commands
 
 ### Development
 
-Head over to the [Docker Development wiki page](https://github.com/rhubarbgroup/redis-cache/wiki/Docker-Development) to spin up various Redis setups.
```
tillkruss, line 51: We should move this to the Wiki page. Do you have permission to update it?

tillkruss (changes requested):

README.md
tillkruss: Sure 👍

tests/TestCase.php
tillkruss: I meant the array 🙃

zaherg: the new line after `global` add more readability to it, so win/win

PR #367, opened 2022-07-19
Add is_predis method

### LINKED ISSUE #366: Add isPredis() : bool
There are several `$this->redis instanceof Predis\Client` checks in the code.

Would you welcome an `isPredis` method?

szepeviktor:
... instead of `$this->redis instanceof Predis\Client`

@tillkruss Here we go.

Closes #366


tillkruss:
Beauty!

Issue #369, opened 2022-07-20
Undefined function


szepeviktor:
`$args = is_predis()`

Does PHPStan spot this?

szepeviktor:
We do not use PHPStan :(
https://github.com/rhubarbgroup/redis-cache/blob/f9c4162558fd237425066b93214fde3817fb7bcb/composer.json#L21-L35

tillkruss:
@zaherg: Wanna set up low level phpstan?

szepeviktor:
@zaherg There are forgotten seeds in https://github.com/rhubarbgroup/redis-cache/pull/298

tillkruss:
@zaherg: Let's run these in GHA as well.

szepeviktor:
There is at least one grave error: includes/class-plugin.php:207
https://developer.wordpress.org/reference/functions/wp_add_dashboard_widget/

zaherg:
@tillkruss @szepeviktor sure thing, it look like it slipped out when I was working on the local setup.

I'll open a PR soon and hopefully @szepeviktor can help guiding me to complete it.

@tillkruss should I create a separate GHA file for it call it `phpstan.yml` or just add it to the current one?

PR #371, opened 2022-07-20
Adding back the missing PHPStan implementation

zaherg:
As mentioned here https://github.com/rhubarbgroup/redis-cache/issues/369 when I implemented the new enhanced implementation for testing, I removed the code related to PHPStan.

This PR should add PHPStan back while also provide the ability to auto run it using Github actions.

cc @szepeviktor your feedback is appreciate it before I make it as ready for review.


zaherg:
@szepeviktor I just wonder, if you don't mind answering, why did you added the `sed` and changed the class name, is there a specific reason for that?

szepeviktor:
Please fix includes/class-plugin.php:207 first. That is a critical error.

zaherg:
got it, thanks for the explanation.

currently, phpstan reports 77 issues at level 5. so it looks like there are few things to fix.

szepeviktor:
> currently, phpstan reports 77 issues

Most of them take 20 seconds to fix.

szepeviktor:
> > Please fix includes/class-plugin.php:207 first. That is a critical error.
> 
> you mean those two errors?

Yes.

zaherg:
based on the [link](https://developer.wordpress.org/reference/functions/wp_add_dashboard_widget/), two are missing.

just pushed a fix

szepeviktor:
> just pushed a fix

You are almost ready.

zaherg:
> > just pushed a fix
> 
> You are almost ready.

Am I missing anything else?

szepeviktor:
> Am I missing anything else?

No, the last commit is complete.

There is only half an hour work and the project will be error-free.

zaherg:
I'll have to look at it later, its midnight here so its time for me to 😴

szepeviktor:
WordPress core is not PHP 8.1 compatible.
The WordPress ecosystem is not PHP 8.0 compatible.
Actually I run every WP installation on PHP 7.4

Please consider running PHPStan on 8.0
I'm not sure!

szepeviktor:
# 🚀

includes/class-plugin.php
```diff
@@ -282,7 +282,7 @@ public function enqueue_admin_styles() {
         if ( ! in_array( $screen->id, $screens, true ) ) {
             return;
         }
-
+        /** @phpstan-ignore-next-line */
         wp_enqueue_style( 'redis-cache', WP_REDIS_PLUGIN_DIR . '/assets/css/admin.css', null, WP_REDIS_VERSION );
```
szepeviktor, line 286: @zaherg Could an empty array in place of `null` fix this?
https://developer.wordpress.org/reference/functions/wp_enqueue_style/

zaherg: just pushed the fix, I was reading the error from `wp_enqueue_script` while updating it 🫣 , sorry about that, WP is not my strongest feature

zaherg:
I'll skip the CI from now on until I can confirm that I was able to fix the one I know how to fix

szepeviktor:
> I'll skip the CI from now on

GHA won't run if you add `[skip ci]` to the commit message.

zaherg:
> > I'll skip the CI from now on
> 
> GHA won't run if you add `[skip ci]` to the commit message.

This is what I was going to do with the future commits

zaherg:
@szepeviktor do you like to have all the fix (the one I can) in one PR or we can now push this small fix, and then work on the others one by one, so I don't break the code ?

szepeviktor:
Would you like me to send a PR to this branch resolving all problems reported?
It takes only couple of minutes.

zaherg:
> Would you like me to send a PR to this branch resolving all problems reported? It takes only couple of minutes.

If you can, thats would be awesome

szepeviktor:
> If you can, thats would be awesome

🏃🏻‍♂️

szepeviktor:
Almost ready.

Please bring in Relay stubs.

zaherg:
ok, one min

szepeviktor:
Please review https://github.com/zaherg/redis-cache/pull/1

szepeviktor:
> Found 23 errors

🎉

szepeviktor:
Please spend 5 more minutes fixing some of those.
Then put all the remaining errors into a baseline file.

zaherg:
I'll do it later today .. thanks @szepeviktor for all the help

includes/ui/tabs/overview.php
```diff
@@ -7,6 +7,8 @@
 
 defined( '\\ABSPATH' ) || exit;
 
+/** @var \Rhubarb\RedisCache\Plugin $roc */
+
```
tillkruss, line 11: Let's drop this space.

phpstan.neon.dist
```diff
@@ -1,22 +1,33 @@
...
+    scanFiles:
+        - php-tools/relay.stub.php
+
     paths:
         - includes/
         - php-tools/object-cache.php
-    excludes_analyse:
+
+    excludePaths:
         - includes/object-cache.php
```
tillkruss, line 29: Excluding `includes/object-cache.php` seems risky, @szepeviktor?

includes/object-cache.php
```diff
@@ -2472,7 +2473,7 @@ public function switch_to_blog( $_blog_id ) {
             return false;
         }
 
-        $this->blog_prefix = $_blog_id;
+        $this->blog_prefix = (string) $_blog_id;
```
tillkruss, line 2476: I rather keep this a `(int)` across the codebase.

includes/diagnostics.php
```diff
@@ -9,6 +9,8 @@
 
 global $wp_object_cache;
 
+/** @var \Rhubarb\RedisCache\Plugin $roc */
+
```
tillkruss, line 13: Let's drop the extra line.

tillkruss (changes requested):

phpstan.neon.dist
zaherg: Its the same as `php-tools/object-cache.php`, when we run `analyze` we create a copy of `includes/object-cache.php` do some magic (ref here https://github.com/rhubarbgroup/redis-cache/pull/371#issuecomment-1190724245) and run the analyze on it.

includes/ui/tabs/overview.php
szepeviktor: This "defines" `$roc`

I leave it 100% up to you.

includes/object-cache.php
szepeviktor: Okay! Much better!
Than please change type of `blog_prefix` property.

includes/diagnostics.php
szepeviktor: This "defines" $roc

I leave it 100% up to you.

includes/ui/tabs/overview.php
szepeviktor: When it touches the next line: there will be Confusion.

szepeviktor:
"php-tools" was coined by me when PHPStan needed a higher PHP version than `redis-cache`

Now we keep only the renamed class file in that directory.
Could someone come up with a good name??

zaherg:
> "php-tools" was coined by me when PHPStan needed a higher PHP version than `redis-cache`
> 
> Now we keep only the renamed class file in that directory. Could someone come up with a good name??

we can change it to `.includes` or `.phpstan`? as it is only used only within PHPStan

szepeviktor:
People with more experience call this directory `build`

zaherg:
I don't think `build` is the correct wording here, but for me any wording works, at the end its not part of the core code, but I'll leave it up to you and Till to decide what to call it.

tillkruss:
@zaherg: Let me know when you're done and I'll take a look at the remaining issues.

zaherg:
@tillkruss i don't think I have anything else to add to it.

PR #372, opened 2022-07-25
Fix few more PHPStan errors

szepeviktor:
Please see comments.

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -771,7 +771,7 @@ class Plugin {
 
                 $messages = get_settings_errors( 'redis-cache' );
 
-                if ( ! empty( $messages ) ) {
+                if ( 0 !== count( $messages ) ) {
                     set_transient( 'settings_errors', $messages, 30 );
 
                     wp_safe_redirect(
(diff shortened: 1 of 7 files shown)


includes/class-plugin.php
```diff
@@ -916,7 +917,6 @@ public function maybe_print_comment() {
         }
 
         if (
-            ! isset( $wp_object_cache->cache_hits ) ||
```
szepeviktor, line 919: This is an integer. Defaults to 0.

includes/object-cache.php
```diff
@@ -1357,7 +1358,7 @@ protected function add_or_replace( $add, $key, $value, $group = 'default', $expi
      * @param   string $group      The group value appended to the $key.
      * @return  bool               Returns TRUE on success or FALSE on failure.
      */
-    public function delete( $key, $group = 'default' ) {
+    public function delete( $key, $group = 'default', $deprecated = false ) {
```
szepeviktor, line 1361: PHP needs all parent parameters.

szepeviktor:
Who is willing to shorten the  baseline file? 😉

includes/object-cache.php
szepeviktor: Or am I wrong?

This error somehow cannot be cannot ignored ...

PR #376, opened 2022-07-28
Bug with info() and ping() when used RedisCluster and named configuration

alexva24:


tillkruss:
Is this a followup to #362?

alexva24:
yes

includes/object-cache.php
```diff
@@ -514,11 +514,18 @@ public function __construct( $fail_gracefully = true ) {
...
-                    : $this->redis->ping( $connectionID );
+	            if ( is_string( WP_REDIS_CLUSTER ) ) {
+		            $connectionID = WP_REDIS_CLUSTER;
+	            } else {
+		            $connectionID = current( $this->build_cluster_connection_array() );
+	            }
+
+	            if ( $client === 'predis' ) {
+		            $this->diagnostics[ 'ping' ] = $this->redis->getClientFor( $connectionID )->ping();
+	            } elseif ( $client === 'phpredis' ) {
```
tillkruss, line 526: This condition break Credit/Relay compatibility, just use an `else`, or keep the old format but fix the `$connectionID`.

includes/object-cache.php
```diff
@@ -1056,11 +1063,15 @@ public function fetch_info() {
...
+            $client = $this->determine_client();
+
+            if ( $client === 'predis' ) {
+                $connectionID = current( $this->build_cluster_connection_array() );
 
-            $info = $this->determine_client() === 'predis'
-                ? $this->redis->getClientFor( $connectionID )->info()
-                : $this->redis->info( $connectionID );
+                $info = $this->redis->getClientFor( $connectionID )->info();
+            } elseif ( $client === 'phpredis' ) {
```
tillkruss, line 1072: Same here. Just use the old format but add your `$connectionID` variable.

tillkruss (changes requested):

includes/object-cache.php
alexva24: I made changes, but if constant WP_REDIS_CLUSTER is defined Exception will thrown here https://github.com/rhubarbgroup/redis-cache/blob/develop/includes/object-cache.php#L726

includes/object-cache.php
```diff
@@ -1056,11 +1058,16 @@ public function fetch_info() {
...
+            $client = $this->determine_client();
+	        $connectionID = current( $this->build_cluster_connection_array() );
 
-            $info = $this->determine_client() === 'predis'
-                ? $this->redis->getClientFor( $connectionID )->info()
-                : $this->redis->info( $connectionID );
+	        if ( $client === 'predis' ) {
+                $info = $this->redis->getClientFor( $connectionID )->info();
+            } elseif ( $client === 'phpredis' ) {
+                $info = $this->redis->info( 'SERVER' );
```
tillkruss, line 1067: @alexva24: Final question, I see that the only big change here is that we're calling `RedisCluster::info( 'SERVER' );` instead of `RedisCluster::info();`. What exact error message is this fixing?

tillkruss: The PhpRedis syntax is:

```php
public function info(string|array $node, ?string $section = null): array;
```

So I don't think this works?

tillkruss (changes requested):

includes/object-cache.php
alexva24: **The method RedisCluster::info() signature is:**
* @param   string $option Optional. The option to provide redis.
*                         SERVER | CLIENTS | MEMORY | PERSISTENCE | STATS | REPLICATION | CPU | CLASTER | KEYSPACE  | COMANDSTATS
    _public function info($option = null) {}_

Using info( 'SERVER' ) reduced count of returning data. See [https://redis.io/commands/info/#notes](https://redis.io/commands/info/#notes)

I tested my changes and already use it my production server.

alexva24: Error message"Warning RedisCluster::info(): Directed commands must be passed a key …
…or [host,port] array"

tillkruss: https://github.com/phpredis/phpredis/blob/677c9da3cbff53b74f7b26ca7dcbdf7e37e1da04/redis_cluster.stub.php#L165

That's the correct behavior.

alexva24: So, i made some changes

PR #378, opened 2022-08-02
Replication bug

lhmook:
true not work..
true => 'predis'

AND 
wp-config.php
WP_REDIS_SERVERS
parameter : alias=master does not work. role=master & role=replica
ex) define( 'WP_REDIS_SERVERS',['tcp://127.0.0.1:6379?database=0&role=master','tcp://127.0.0.1:6380?database=0&role=replica']);


tillkruss:
Thank you, this changed in Predis v2.0

PR #381, opened 2022-08-05
fix for when cache item's value is set to null

nickchomey:
If a cache item has a value of "null", the current isset() mechanism returns false. This results in the cache key never being used and, in my case, hundreds of duplicate queries on each page load. This changes it to array_key_exists. Likewise, it would never delete the cache key, and probably other issues.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1340,7 +1340,7 @@ class WP_Object_Cache {
             }
         }
 
-        $exists = isset( $this->cache[ $derived_key ] );
+        $exists = array_key_exists( $derived_key, $this->cache );
 
         if ( (bool) $add === $exists ) {
             return false;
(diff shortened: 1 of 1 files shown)


tillkruss:
I'm happy to merge this, but IMO storing `null` or `false` as a cache value is silly, because you can't tell if they key didn't exist or was `null|false`.

tillkruss:
@nickchomey Would you mind sharing what plugin is storing `null` values?

tillkruss:
Ah I see. I guess I'd store `0`. I'm gonna write test for this next week for Object Cache Pro and make sure it handles this well if it doesn't already.

tillkruss:
Link it here and I'll push it via Slack 👍

tillkruss:
No, I've got a Slack channel with them for Object Cache Pro. I'll try to promote the PR.

tillkruss:
Yeah, Object Cache Pro already supports `null` values, but I added tests just in case.

These are quite risky values, because Predis returns `null` when a key doesn't exists and PhpRedis returns `false` (by default), so storing `null|false` often leads to issues.

Anyhow since WP's API expects `false` I reckon supporting `null` is somewhat justified.

Screen Shot 2022-08-08 at 1 47 01 PM

tillkruss:
Annoying. I'll tag a release shortly.

Issue #379, opened 2022-08-05
get_multiple() can cause an "Invalid argument supplied for foreach()" warning


ocean90:
I'm seeing a few warnings in our Sentry data where `$results` in https://github.com/rhubarbgroup/redis-cache/blob/afe483c4b0405fc997bd2bb89cf44350a40093b6/includes/object-cache.php#L1909 is not set, causing the "Warning: Invalid argument supplied for foreach()" PHP warning.
It seems like this happens when [the previous try/catch block](https://github.com/rhubarbgroup/redis-cache/blob/afe483c4b0405fc997bd2bb89cf44350a40093b6/includes/object-cache.php#L1885-L1902) catches an exception which sets `$cache` but not `$results`. ~Is it as simple as renaming the variable?~

I currently have no details about the exact exception which is thrown there but I'll follow-up on it once I have it.

<details>
<summary>Example stack backtraces</summary>

```
ErrorException: Warning: Invalid argument supplied for foreach()
#22 /home/vcap/app/web/content/object-cache.php(1904): WP_Object_Cache::get_multiple
#21 /home/vcap/app/web/content/object-cache.php(171): wp_cache_get_multiple
#20 /wp-includes/functions.php(6991): _get_non_cached_ids
#19 /wp-includes/taxonomy.php(3989): _prime_term_caches
#18 /wp-includes/taxonomy.php(3659): get_object_term_cache
#17 /wp-includes/post.php(7797): wp_queue_posts_for_term_meta_lazyload
#16 /wp-includes/class-wp-query.php(3344): WP_Query::get_posts
#15 /wp-includes/class-wp-query.php(3586): WP_Query::query
#14 /wp-includes/post.php(2410): get_posts
#13 /home/vcap/app/web/content/themes/theme/inc/template-functions.php(135): theme_add_feed_item_media
#12 /wp-includes/class-wp-hook.php(307): WP_Hook::apply_filters
#11 /wp-includes/class-wp-hook.php(331): WP_Hook::do_action
#10 /wp-includes/plugin.php(476): do_action
#9 /wp-includes/feed-rss2.php(125): require_once
#8 /wp-includes/template.php(770): load_template
#7 /wp-includes/functions.php(1662): do_feed_rss2
#6 /wp-includes/class-wp-hook.php(309): WP_Hook::apply_filters
#5 /wp-includes/class-wp-hook.php(331): WP_Hook::do_action
#4 /wp-includes/plugin.php(476): do_action
[...]

PR #383, opened 2022-08-08
When retrieving multiple values ensure the results variable is properly set in case of an exception

ocean90:
Related to #379.

`$cache` should be filled based on `$results`.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1881,14 +1881,14 @@ LUA;
         $start_time = microtime( true );
         $results = [];
 
-        try {
-            $remaining_ids = array_map(
-                function ( $key ) use ( $derived_keys ) {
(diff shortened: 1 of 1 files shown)


includes/object-cache.php
```diff
@@ -1897,7 +1897,10 @@ function ( $key ) use ( $derived_keys ) {
         } catch ( Exception $exception ) {
             $this->handle_exception( $exception );
 
-            $cache = array_fill( 0, count( $derived_keys ) - 1, false );
```
ocean90, line 1900: fwiw: this currently drops any cache hits from the internal cache set in https://github.com/rhubarbgroup/redis-cache/blob/c70718a9bfa3c87a154f6b6a38125c4275cd1b34/includes/object-cache.php#L1857-L1867.

PR #385, opened 2022-08-24
Fix array_map error with empty result of exec method

tkalimullin-coludlinux:
Received an error when calling the `Redis->pipeline->exec` method, which may not return an array, resulting in the following log:

```
PHP Warning:  array_map(): Expected parameter 2 to be an array, bool given in /home/user/public_html/wp-content/object-cache.php on line 1212
PHP Warning:  array_combine() expects parameter 2 to be array, null given in /home/user/public_html/wp-content/object-cache.php on line 1214
PHP Warning:  Invalid argument supplied for foreach() in /home/user/public_html/wp-content/object-cache.php on line 1216
```

Added a result check to make sure we got an array to pass to the array_map function.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1210,9 +1210,15 @@ class WP_Object_Cache {
 
             $method = $this->is_predis() ? 'execute' : 'exec';
 
+            $results = $tx->{$method}();
+
+            if ( ! is_array( $results ) ) {
+                $results = [];
+            }
+
             $results = array_map( function ( $response ) {
                 return (bool) $this->parse_redis_response( $response );
-            }, $tx->{$method}() );
+            }, $results );
 
             $results = array_combine( $keys, $results );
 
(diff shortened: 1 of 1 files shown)


naxvog:
Wouldn't a simple `$tx->{$method}() ?? []` be enough?
Don't know for sure but this might be a micro optimization in itself.

tkalimullin-coludlinux:
I've noticed phpcs.xml, where testing versions of php start at 5.6, and the join operator `??` was added from version 7.

tkalimullin-coludlinux:
[naxvog](https://github.com/naxvog) but I'll agree with you because the plugin is available since version 7.2, thanks. Pushed.

tillkruss:
Thanks @tkalimullin-coludlinux!

tillkruss:
@tkalimullin-coludlinux This didn't actually solve it, see 7361b29.

```php
false ?? [] // false
```

PR #391, opened 2022-09-02
Really fix all EC problems

szepeviktor:
@tillkruss #390 was half-done and ❌failing.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -20,3 +20,6 @@ indent_size = 2
 
 [LICENSE.md]
 indent_style = unset
+
+[phpstan-baseline.neon]
+indent_style = tab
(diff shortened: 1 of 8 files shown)


tillkruss:
The `readme.txt` changes won't work like that.

szepeviktor:
Yes. Wrong `sed` command!

szepeviktor:
@tillkruss I'd like to with those few PHPStan ignores...

szepeviktor:
Now everything should be 🍏

szepeviktor:
Very strange! On my server "Ignored error pattern ... was not matched"
In CI it is all green!

szepeviktor:
> Very strange! On my server "Ignored error pattern ... was not matched",
> in CI it is all green!

Got it! The cause is the latest release of PHPStan: v1.8.3

PR #392, opened 2022-09-12
Throw exception when Redis pipeline returns unexpected results

tkalimullin-coludlinux:
If the Redis pipeline returns an unexpected result, we will get an error when working with the array_combine function:
"Warning: array_combine(): Both parameters should have an equal number of elements"

To avoid a false positive, it's best to stop interacting with the cache.


tillkruss:
Turns out the stubs are not strict: https://github.com/phpredis/phpredis/pull/2015

tillkruss:
I wonder if this comes from `wp_cache_add_multiple([])` calls or something weird.

tillkruss:
Related 7361b29.

PR #394, opened 2022-09-24
Add support for the new storage classes in Query Monitor

johnbillion:
The next release of Query Monitor will [switch to using a `QM_Data` instance for the data storage for each collector](https://github.com/johnbillion/query-monitor/pull/671). The class implements `ArrayAccess` so there's generally no need for third party collectors to change their behaviour, except in the case where they try to write to a second-level element within an array in which case an "Indirect modification of overloaded element of QM_Data_Fallback has no effect" PHP notice will be triggered because it's not possible to write to a property of an overloaded element.

Without this fix the code still works, it just doesn't report the figures for hits and misses. With this fix, all is well again.

### DIFF
--- a/includes/class-qm-collector.php
+++ b/includes/class-qm-collector.php
@@ -73,8 +73,13 @@ class QM_Collector extends Base_Collector {
         ];
 
         // These are used by Query Monitor.
-        $this->data['stats']['cache_hits'] = $info->hits;
-        $this->data['stats']['cache_misses'] = $info->misses;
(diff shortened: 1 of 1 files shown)


tillkruss:
Thanks John!

tillkruss:
@johnbillion: I added a followup commit: https://github.com/rhubarbgroup/redis-cache/commit/826da7b8a20a363f7d0466aab89d8d0816686890

PR #398, opened 2022-11-10
Add wp_cache_flush_group implementation

wilburpowery:
resolves #396

- Add wp_cache_flush_group implementation
- Rename method

PR #399, opened 2022-11-10
Add wp_cache_flush_group implementation

wilburpowery:


wilburpowery:
Duplicate of #398

includes/object-cache.php
```diff
@@ -1629,6 +1642,75 @@ public function flush( $delay = 0 ) {
...
+    {
+        $san_group = $this->sanitize_key_part( $group );
+
+        if ( is_multisite() && ! $this->is_global_group( $san_group ) ) {
+            $salt = str_replace( "{$this->blog_prefix}:", '*:', $this->fast_build_key( '*', $san_group ) );
+        } else {
+            $salt = $this->fast_build_key( '*', $san_group );
+        }
+
+        $this->cache = []; // TODO: remove all matching keys from  $this->cache
```
tillkruss, line 1661: This is looking good, just needs to iterate over `$this->cache` and delete all keys that match the pattern. Because this is internal we can just use:

```php
strpos($id, "{$san_group}:") === 0 || strpos($id, ":{$san_group}:") !== false
```

tillkruss (changes requested):

PR #406, opened 2023-01-09
Fix grokzen/redis-cluster Docker image

szepeviktor:
v5 was dropped

https://github.com/Grokzen/docker-redis-cluster/commit/d597529abb5d9e8b02721c7a2a690f7100fd67e8
https://hub.docker.com/r/grokzen/redis-cluster/tags

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -111,7 +111,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ['7.2', '7.3', '7.4', '8.0', '8.1']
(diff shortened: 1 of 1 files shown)


tillkruss:
Can we do a 4.x or 5.x version as well?

szepeviktor:
> Can we do a 4.x or 5.x version as well?

No, these Docker tags do not exists any more. Please see the above link.

szepeviktor:
Should we release those old version here on GitHub Container Registry?

PR #408, opened 2023-01-09
Improve static analysis

szepeviktor:
~~This is just a test ⚠️~~

### DIFF
--- a/includes/diagnostics.php
+++ b/includes/diagnostics.php
@@ -102,15 +102,16 @@ foreach ( $constants as $constant ) {
 }
 
 if ( defined( 'WP_REDIS_PASSWORD' ) ) {
+    /** @var string|array|null $password */
     $password = WP_REDIS_PASSWORD;
 
     if ( is_array( $password ) ) {
-        if ( isset( $password[1] ) && ! is_null( $password[1] ) && '' !== $password[1] ) {
+        if ( isset( $password[1] ) && '' !== $password[1] ) {
             $password[1] = str_repeat( '•', 8 );
         }
 
         $info['WP_REDIS_PASSWORD'] = wp_json_encode( $password, JSON_UNESCAPED_UNICODE );
-    } elseif ( ! is_null( $password ) && '' !== $password ) {
+    } elseif ( is_string( $password ) && '' !== $password ) {
         $info['WP_REDIS_PASSWORD'] = str_repeat( '•', 8 );
     }
 }
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,7 +11,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['8.2']
+        php: ['7.4','8.0','8.1']
+        relay: ['0.5.1']
 
     steps:
       - name: Checkout
@@ -22,6 +23,17 @@ jobs:
         with:
           php-version: ${{ matrix.php }}
           extensions: msgpack, igbinary, mbstring, redis-5.3.7
+
+      - name: Install Relay PHP extension
+        run: |
+          curl -L "https://builds.r2.relay.so/${{ matrix.relay }}/relay-${{ matrix.relay }}-php${{ matrix.php }}-debian-x86-64.tar.gz" | tar xz
+          cd relay-${{ matrix.relay }}-php${{ matrix.php }}-debian-x86-64/
+          sudo cp -v relay.ini $(php-config --ini-dir)
+          sudo cp -v relay-pkg.so $(php-config --extension-dir)/relay.so
+          sudo sed -i -e "s/00000000-0000-0000-0000-000000000000/$(cat /proc/sys/kernel/random/uuid)/" $(php-config --extension-dir)/relay.so
+
+      - name: Show Relay configuration
+        run: php --ri relay
 
       - name: Install PHP Dependencies
         uses: ramsey/composer-install@v2
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ['7.4','8.0','8.1']
-        relay: ['0.5.1']
+        relay: ['0.4.6','0.5.1']
 
     steps:
       - name: Checkout
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         php: ['7.4', '8.2']
-        relay: ['0.4.6', '0.5.1']
+        relay: ['0.5.1']
 
     steps:
       - name: Checkout
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,6 +23,7 @@ jobs:
         with:
           php-version: ${{ matrix.php }}
           extensions: msgpack, igbinary, mbstring, redis-5.3.7
+          coverage: none
 
       - name: Install Relay PHP extension
         run: |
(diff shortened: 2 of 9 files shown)


szepeviktor:
@tillkruss Does `Relay` support only PHP 8.2?

builds.r2.relay.so/dev/relay.stub.php

tillkruss:
No, 7.4 — 8.2.

szepeviktor:
Could you release a stub for PHP 7.4?

tillkruss:
No, extension stubs are always generated using php `stable`.

szepeviktor:
Okay. Current stubs are syntactically incorrect: empty functions. So they are currently (in this PR) not processed.

> PHP Fatal error:  Non-abstract method Relay\Relay::__construct() must contain body in relay.stub.php on line 433

And PHPStan is not able to discover new classes/methods in stubs.

szepeviktor:
> What to do now?

It turns out adding `{}` help.
```php
    public function __construct(
        string $host = null,
        int $port = 6379,
        float $connect_timeout = 0.0,
        float $command_timeout = 0.0,
        #[\SensitiveParameter] array $context = [],
        int $database = 0,
    ) {}
```

Is it possible to generate stubs with `{}`?

tillkruss:
Stubs are auto-generated, no need for syntax checks.

https://github.com/php/php-src/blob/master/build/gen_stub.php

szepeviktor:
Okay. But PHPStan needs **valid** source code for analysis.

szepeviktor:
May I automate adding those `{}` ?

tillkruss:
There is no need to analyze the stubs.

You can still load them regardless: https://phpstan.org/user-guide/stub-files

szepeviktor:
On that page I read

> Stub files aren’t a replacement for [discovering symbols](https://phpstan.org/user-guide/discovering-symbols) so if you’re trying to fix errors like “Function not found” or “Class not found”, check out the [discovering symbols](https://phpstan.org/user-guide/discovering-symbols) guide instead.

that is why we have an error: "Instantiated class Relay\Relay not found."

tillkruss:
I see. In that case do whatever you like. This is what I do for Object Cache Pro:

```yaml
parameters:
    level: 7
    scanFiles:
        - tests/PHPStan/relay.stub.php
```

szepeviktor:
> ```yaml
>     scanFiles:
>         - tests/PHPStan/relay.stub.php
> ```

That should also fail ⚡ 
`scanFiles` expects valid PHP code.
Don't you have undefined `Relay\Relay` errors?

tillkruss:
No passes for me.

szepeviktor:
> No passes for me.

Does it pass without stubs?

tillkruss:
That could be it, why don't you install it here too?

https://relay.so/docs/1.x/installation#github-actions

szepeviktor:
> That could be it, why don't you install it here too?

You are right!!

1. PHPStan works without stubs
2. stubs fail in Object Cache Pro too

szepeviktor:
@tillkruss Look 👶🏻 who's talking!

szepeviktor:
@tillkruss 🚨 Somehow PHP union type got into Relay 0.5.1 for PHP 8.0 😿 
```php
public function exists(mixed ...$keys) : \Relay\Relay|int|bool
    {
    }
    public function eval(mixed $script, array $args = [], int $num_keys = 0) : ?\Relay\Relay|mixed
    {
    }
    public function eval_ro(mixed $script, array $args = [], int $num_keys = 0) : ?\Relay\Relay|mixed
    {
    }
```
Could you have that fixed? Viktor has no access ...

tillkruss:
No, like I said, stubs are generated by PHP itself, according to their standards, we don't post-process them.

szepeviktor:
> stubs are generated by PHP itself

Here we have no more stubs. PHPStan runs well on PHP 7.4 and 8.1 but not on 8.0.
And shows the above stub-like three functions.

szepeviktor:
@tillkruss Please merge this one.

The error on PHP 8.0 will be debugged in the next PR.

szepeviktor:
@tillkruss Now as my server is full of Relay :) I'm sorry to report but the the output of PHPStan: "eval_ro" is nowhere to find.

- not in redis-cache
- not in dependencies
- not in PHPStan PHAR file

😲 but in relay.so

So the only file PHPStan can read `eval_ro` is the extension binary.
```php
public function eval_ro(mixed $script, array $args = [], int $num_keys = 0) : ?\Relay\Relay|mixed
```
Then the union type must come from the .so file.

If Viktor would have access the C source code then he could dig deeper.

szepeviktor:
# 💣

But `php --re` tells use "Till's secret"
```
        Method [ <internal:relay> public method eval_ro ] {

          - Parameters [3] {
            Parameter #0 [ <required> mixed $script ]
            Parameter #1 [ <optional> array $args = [] ]
            Parameter #2 [ <optional> int $num_keys = 0 ]
          }
          - Return [ Relay\Relay|mixed ]
        }
```
There it is: union return type.

Could you have your guys fix int?

szepeviktor:
@tillkruss I think I got it.

1. `mixed` means any type, as PHP error message says: _Type mixed can only be used as a standalone type_
2. nullable type is not `?\Relay\Relay|mixed` but `\Relay\Relay|mixed|null`

So where mixed is in a union it should be simply `mixed`.
Could you get your guys implement that?

tillkruss:
Yeah that might actually be a bug. Let me confirm.

tillkruss:
Use this stub: https://builds.r2.relay.so/dev/relay.stub.php

szepeviktor:
I'm sorry to report that this stubs file is not PHP 8.0 compatible :(

szepeviktor:
We have received a sentence from Ondrey https://github.com/phpstan/phpstan/blob/bdf6dbc3e7ca2745d2f7256d83916c1e734ad3df/website/src/user-guide/troubleshooting-types.md#internal-stubs

> For custom PHP extensions that are not present in jetbrains/phpstorm-stubs, their [PHP reflection data](https://www.php.net/manual/en/book.reflection.php) is used.

tillkruss:
For context.

Screenshot 2023-01-10 at 3 28 40 PM

szepeviktor:
Thank you.
The above problems are not only in the stubs bug in the extensions Reflections as Ondrey wrote.

szepeviktor:
> drop 8.0 checks

Accidentally 8.1 was dropped too.

szepeviktor:
> gzip: stdin: not in gzip format

Relay for 8.2 is not yet published in the R2 bucket.

szepeviktor:
@tillkruss I have a hunch that I'm going to find new problem/bugs.

Could we make a up (write a document about it) a PHP/Relay version matrix that is supported?
I'm available on Slack/phone/by smoke signals 🙃

includes/diagnostics.php
```diff
@@ -102,15 +102,16 @@
...
 
     if ( is_array( $password ) ) {
-        if ( isset( $password[1] ) && ! is_null( $password[1] ) && '' !== $password[1] ) {
+        if ( isset( $password[1] ) && '' !== $password[1] ) {
             $password[1] = str_repeat( '•', 8 );
         }
 
         $info['WP_REDIS_PASSWORD'] = wp_json_encode( $password, JSON_UNESCAPED_UNICODE );
-    } elseif ( ! is_null( $password ) && '' !== $password ) {
+    } elseif ( is_string( $password ) && '' !== $password ) {
```
tillkruss, line 114: This better not fuck us over

tillkruss (approved):

tillkruss:
> @tillkruss I have a hunch that I'm going to find new problem/bugs.

🫠

tillkruss:
> Could we make a up (write a document) a PHP/Relay version matrix that is supported?

```
>=0.5.1 (PHP 7.4 — 8.2)
>=0.2.1 (PHP 7.4 — 8.1)
>=0.1.0 (PHP 7.4 — 8.0)
```

PR #418, opened 2023-02-08
Admin bar Object Cache menu

### LINKED ISSUE #224: "Object Cache" admin-bar node
How can this be the most useful?

- [ ] ~~Toggle via options in settings~~
- [x] Enabled by default?
- [x] Disabled via `WP_REDIS_DISABLE_ADMINBAR`
- [x] Add with priority of 998, just before Query Monitor? 
- [ ] Should we hide the admin dashboard as well via an option?

The node could be:

```
| Object Cache
| - Flush Cache
| - Show Metrics
| --------------
| Status: Connected
| 99.4%; 940/6; 98.09 KB
```

Related:

- https://onexa.nl/wordpress/toolbar-link-redis-object-cache/?v=3e8d115eb4b3
- https://wordpress.org/support/topic/fyi-flush-button-script-breaks-update/#post-13121765

A5hleyRich:
Resolves #224 

I did consider removing the "Show Metrics" node and making the actual metrics clickable. Thoughts?

CleanShot 2023-02-08 at 14 24 05@2x


includes/class-plugin.php
```diff
@@ -643,6 +645,82 @@ public function show_admin_notices() {
...
+                ]
+            );
+        }
+
+        if ( Metrics::is_enabled() ) {
+            $wp_admin_bar->add_node(
+                [
+                    'parent' => 'redis-cache',
+                    'id' => 'redis-cache-metrics',
+                    'title' => __( 'Show Metrics', 'redis-cache' ),
```
tillkruss, line 687: Should we make the "Settings"?

includes/class-plugin.php
```diff
@@ -643,6 +645,82 @@ public function show_admin_notices() {
...
+            global $wp_object_cache;
+
+            $info = $wp_object_cache->info();
+
+            $wp_admin_bar->add_node(
+                [
+                    'parent' => 'redis-cache-info',
+                    'id' => 'redis-cache-info-details',
+                    'title' => sprintf(
+                        '%s%%&nbsp;&nbsp;%d/%d&nbsp;&nbsp;%s',
```
tillkruss, line 713: Can we make this a link to the metrics screen?

includes/class-plugin.php
```diff
@@ -643,6 +645,82 @@ public function show_admin_notices() {
...
+                        $info->ratio,
+                        $info->hits,
```
tillkruss, line 715: This might be `52,234` we should format it, so its readable.

tillkruss (changes requested):

includes/class-plugin.php
A5hleyRich: To confirm. This should be renamed "Settings" and dump them here:

CleanShot 2023-02-08 at 19 22 34@2x

tillkruss: Yeah 👍

includes/class-plugin.php
```diff
@@ -643,6 +645,92 @@ public function show_admin_notices() {
...
+                    'href' => $this->action_link( 'flush-cache' ),
+                ]
+            );
+        } else {
+            $wp_admin_bar->add_node(
+                [
+                    'parent' => 'redis-cache',
+                    'id' => 'redis-cache-enable',
+                    'title' => __( 'Enable Cache', 'redis-cache' ),
+                    'href' => $this->action_link( 'enable-cache' ),
```
tillkruss, line 686: Let's send them to settings instead of directly calling the action.

includes/class-plugin.php
```diff
@@ -643,6 +645,83 @@ public function show_admin_notices() {
...
+                [
+                    'parent' => 'redis-cache',
+                    'id' => 'redis-cache-flush',
+                    'title' => __( 'Flush Cache', 'redis-cache' ),
+                    'href' => $this->action_link( 'flush-cache' ),
+                ]
+            );
+        }
+
+        if ( Metrics::is_enabled() ) {
```
tillkruss, line 682: This seem not necessary, eh?

includes/class-plugin.php
```diff
@@ -643,6 +645,83 @@ public function show_admin_notices() {
...
+            return;
+        }
+
+        if ( ! $this->current_user_can_manage_redis() ) {
+            return;
+        }
+
+        $wp_admin_bar->add_node(
+            [
+                'id'    => 'redis-cache',
```
tillkruss, line 666:
```suggestion
                'id' => 'redis-cache',
```

includes/class-plugin.php
```diff
@@ -643,6 +645,83 @@ public function show_admin_notices() {
...
+            ]
+        );
+
+        if ( $this->get_redis_status() ) {
+            $wp_admin_bar->add_node(
+                [
+                    'parent' => 'redis-cache',
+                    'id' => 'redis-cache-flush',
+                    'title' => __( 'Flush Cache', 'redis-cache' ),
+                    'href' => $this->action_link( 'flush-cache' ),
```
tillkruss, line 677: What happens when this is called on the frontend?

includes/class-plugin.php
```diff
@@ -643,6 +645,83 @@ public function show_admin_notices() {
...
+            [
+                'id' => 'redis-cache-info',
+                'parent' => 'redis-cache',
+                'meta' => [
+                    'class' => 'ab-sub-secondary',
+                ],
+            ]
+        );
+
+        if ( $this->get_redis_status() ) {
```
tillkruss, line 703: Let's call this only once and re-use the result.

tillkruss (changes requested):

includes/class-plugin.php
A5hleyRich: The cache is flushed, and the user is redirected to the ROC settings screen.

tillkruss: Let's count the days until someone complains about this 🤣

includes/class-plugin.php
```diff
@@ -101,6 +101,8 @@ public function add_actions_and_filters() {
         add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_admin_scripts' ] );
         add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_redis_metrics' ] );
 
+        add_action( 'admin_bar_menu', [ $this, 'render_admin_bar' ], 998 );
```
tillkruss, line 104: PR looks great. Why the high priority?

A5hleyRich: Taken from the original issue: https://github.com/rhubarbgroup/redis-cache/issues/224

> Add with priority of 998, just before Query Monitor?

PR #425, opened 2023-03-02
Fix `$salt` when `$group` ends with site id

aleks-lynx:
Example:

```
$group = "group1"
$this->blog_prefix = "1"

$salt is "\*:group\*:\*" but should be "\*:group1:\*"

```

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1644,7 +1644,7 @@ class WP_Object_Cache {
         $san_group = $this->sanitize_key_part( $group );
 
         if ( is_multisite() && ! $this->is_global_group( $san_group ) ) {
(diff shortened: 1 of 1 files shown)


aleks-lynx:
@tillkruss I called simply `wp_cache_flush_group('group1')` to reproduce this

aleks-lynx:
Basically I had a small test for cache implementation and groups -- I created group1, group2, and group3. Then I started to delete them one by one, and then noticed that I can't delete group3 after group1 for some reason :) So the reason was that `wp_cache_flush_group('group1')` deleted actually all the keys even for group3.

tillkruss:
Aaaaahh, I see. Nice catch!

PR #439, opened 2023-03-24
remove visible passwords from overview and diagnostics when passing passwords via connection string

dustinrue:
In my situation, I am using Redis with Sentinel option enabled in Kubernetes. All sentinels (and Redis) are available at an internal service of "redis" and it is only necessary to define the single instance. However, when using Redis Sentinel, where the sentinel requires a password, the password may be passed via the configuration like this:

```
define( 'WP_REDIS_SERVERS', [
  'redis:26379/?password=sentinel-password'
] );
```

This information is then visible on the overview and diagnostics screen. This PR redacts the information before display. I did not create an associated "Issue" however.

### DIFF
--- a/includes/diagnostics.php
+++ b/includes/diagnostics.php
@@ -117,6 +117,10 @@ if ( defined( 'WP_REDIS_PASSWORD' ) ) {
     }
 }
 
+if ( defined( 'WP_REDIS_SERVERS' ) ) {
+    $info['WP_REDIS_SERVERS'] = preg_replace('/password=(.+[a-zA-Z])(?=&{0,1})/um', 'password=********', $info['WP_REDIS_SERVERS']);
+}
+
 if ( $dropin && ! $disabled ) {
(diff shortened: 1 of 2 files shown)


includes/diagnostics.php
```diff
@@ -117,6 +117,10 @@
     }
 }
 
+if ( defined( 'WP_REDIS_SERVERS' ) ) {
+    $info['WP_REDIS_SERVERS'] = preg_replace('/password=(.+[a-zA-Z])(?=&{0,1})/um', 'password=********', $info['WP_REDIS_SERVERS']);
```
tillkruss, line 121: Can we make this a helper method, so we don't duplicate the logic?

Do any other constants need to be checked?

tillkruss (changes requested):

tillkruss:
Superseded by 302e7de8a8f90c69c59505fa1a828e3b7aa28045. @dustinrue: LMK if that works for you.

Issue #433, opened 2023-03-14
Addition of custom options hook/filters


damsean102:
Hi,

I was just wondering if there was any plans/thought about introducing a hook/filter to allow other developers the ability to define the settings that can be used [here](https://github.com/rhubarbgroup/redis-cache/blob/main/includes/object-cache.php#L635)?

The reason I ask is that a need has arose for us to connect to REDIS via TLS and where our REDIS server is hosted (Heroku) requires us to alter some of the [SSL parameters](https://devcenter.heroku.com/articles/connecting-heroku-redis#connecting-with-predis) (verify_peer & verify_peer_name) which currently isn't possible using the plugin.
I realise this is a feature of [Object Cache Pro](https://objectcache.pro/docs/configuration-options/#tls-options), but I was wondering if it is likely to make it's way to the free version or not?

Thanks in advance,
Sam

tillkruss:
Hey Sam, you can't use filters, because `object-cache.php` is loaded way before you can register them. We can add a new `WP_REDIS_?` constant tho to disable TLS verification.

You'd just need to ensure it works for PhpRedis as well as Predis. They do that differently.

damsean102:
Hi Till,

Thanks for the speedy reply! Yeah good point about the hooks, a constant could work well though. Something like WP_REDIS_TLS or something more generic (WP_REDIS_CONFIG) where we can pass client appropriate parameters into the Redis or Predis classes would be great.

I can spin up a new REDIS instance and test both those is that helps?

Thanks

tillkruss:
Yup sounds good 👍

david-donachie:
And here I am just sending you an email about the exact same issue! We are stuck in the same issue, where we can't connect to Heroku's self-signed Redis because we can't pass those SSL parameters. Since Heroku will force the upgrade by the 24th of April, this is pretty urgent for us.

Anything we can do to help?

tillkruss:
You might be able to send the TLS parameters when using Predis.

david-donachie:
> You might be able to send the TLS parameters when using Predis.

Not sure I understand? Do you mean that the SSL parameters might be read by Predis from somewhere external to the Wordpress configuration?

tillkruss:
As mentioned by email, Object Cache Pro fully supports TLS and configuration of it.

If you want to open a PR with TLS and it's options for Predis+PhpRedis we gladly merge it.

david-donachie:
[...]

PR #450, opened 2023-04-06
Add WP_REDIS_SSL setting, pass to predis and phpredis

david-donachie:
Simple fix to allow a WP_REDIS_SSL setting to pass SSL connection arguments in the case of a TLS service (such as Heroku) that needs additional SSL parameters.

`connect_using_phpredis()` sets an 'ssl' key on it's connection arguments derived from it via `$parameters['ssl']`, `connect_using_predis()` directly uses the `$parameters['ssl']` value.

e.g.
```php
	define('WP_REDIS_SSL', [
		'verify_host' => false,
		'verify_peer' => false,
		'verify_peer_name' => false,
	]);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ The plugin comes with vast set of configuration options.
 | `WP_REDIS_DISABLE_METRICS`           | `false`     | Disables metrics collection and display |
 | `WP_REDIS_DISABLE_BANNERS`           | `false`     | Disables promotional banners |
 | `WP_REDIS_DISABLE_DROPIN_AUTOUPDATE` | `false`     | Disables the drop-in auto-update |
-| `WP_REDIS_TLS`                       |             | TLS connection options for `tls` or `rediss` scheme |
+| `WP_REDIS_SSL_CONTEXT`               | `[]`        | TLS connection options for `tls` or `rediss` scheme |
 
 </details>
 
@@ -98,10 +98,10 @@ define( 'WP_REDIS_HOST', 'master.ncit.ameaqx.use1.cache.amazonaws.com' );
 define( 'WP_REDIS_PORT', 6379 );
 ```
 
-Additional stream connection options for `phpredis` and `predis` connections can be defined using `WP_REDIS_TLS`
+Additional stream connection options for `phpredis`, `relay`, and `predis` connections can be defined using `WP_REDIS_SSL_CONTEXT`
 
 ```php
-define( 'WP_REDIS_TLS', [
+define( 'WP_REDIS_SSL_CONTEXT', [
     'verify_host' => false,
     'verify_host_name' => false
 ]);
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ The plugin comes with vast set of configuration options.
 | `WP_REDIS_DISABLE_METRICS`           | `false`     | Disables metrics collection and display |
 | `WP_REDIS_DISABLE_BANNERS`           | `false`     | Disables promotional banners |
 | `WP_REDIS_DISABLE_DROPIN_AUTOUPDATE` | `false`     | Disables the drop-in auto-update |
+| `WP_REDIS_TLS`                       |             | TLS connection options for `tls` or `rediss` scheme |
 
 </details>
 
@@ -95,6 +96,15 @@ define( 'WP_REDIS_PATH', '/var/run/redis.sock' );
 define( 'WP_REDIS_SCHEME', 'tls' );
 define( 'WP_REDIS_HOST', 'master.ncit.ameaqx.use1.cache.amazonaws.com' );
 define( 'WP_REDIS_PORT', 6379 );
+```
+
+Additional stream connection options for `phpredis` and `predis` connections can be defined using `WP_REDIS_TLS`
+
+```php
+define( 'WP_REDIS_TLS', [
+    'verify_host' => false,
+    'verify_host_name' => false
+]);
 ```
 
 </details>
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ Additional TLS/SSL stream connection options for connections can be defined usin
 ```php
 define( 'WP_REDIS_SSL_CONTEXT', [
     'verify_host' => false,
-    'verify_host_name' => false
+    'verify_host_name' => false,
 ]);
 ```
 
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ define( 'WP_REDIS_HOST', 'master.ncit.ameaqx.use1.cache.amazonaws.com' );
 define( 'WP_REDIS_PORT', 6379 );
 ```
 
-Additional stream connection options for `phpredis`, `relay`, and `predis` connections can be defined using `WP_REDIS_SSL_CONTEXT`
+Additional TLS/SSL stream connection options for connections can be defined using `WP_REDIS_SSL_CONTEXT`:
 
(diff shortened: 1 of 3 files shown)


tillkruss:
Closes #433.

includes/object-cache.php
```diff
@@ -708,6 +709,7 @@ protected function connect_using_phpredis( $parameters ) {
                     $parameters['scheme'],
                     str_replace( 'tls://', '', $parameters['host'] )
                 );
+                $args['ssl'] = $parameters['ssl'];
```
tillkruss, line 712: Have you tested this with PhpRedis? I don't think this will do anything.

The new constant needs to be added to diagnostics.

SSL isn't really a thing anymore it's all TLS now, should we call it that?

tillkruss (changes requested):

includes/object-cache.php
david-donachie: '''Setting name'''
Both predis and phpredis's documentation suggest you pass this as 'SSL', as far as I could see. Since you set $parameters directly from the name of the setting, you'd have to have an extra predis line to set 'ssl' from 'tls'. I can do that, but it seems odd.

Also, since this is only SSL options, not any other possible TLS options, it seems odd to name it as TLS. But I can put something in build_parameters() to rename it if you feel WP_REDIS_TLS is better?

'''PhpRedis'''
phpredis docs: https://github.com/phpredis/phpredis#class-redis — shows the use of the 'ssl' key in the arguments, so I'm not sure why it would do nothing to add it?

I've not tested with phpredis, though, because I don't have it installed.

'''Diagnostics'''
Maybe I'm not understanding the code, but $diagnostics gets merged with $parameters in connect_using_predis(), so it will include the 'ssl' value, and in connect_using_phpredis() the setting of $args['ssl'] happens before you add $args to $diagnostics, so I think it's included in either case? Happy to make changes if I'm wrong and you can point me towards what I need to change.

includes/object-cache.php
```diff
@@ -641,6 +641,7 @@ protected function build_parameters() {
             'timeout',
             'read_timeout',
             'retry_interval',
+            'ssl'
```
tillkruss, line 644: Let's call it `WP_REDIS_TLS`. It also needs to be documented in the README.

includes/object-cache.php
tillkruss: Moving to separate threads.

includes/object-cache.php
```diff
@@ -708,6 +709,7 @@ protected function connect_using_phpredis( $parameters ) {
                     $parameters['scheme'],
                     str_replace( 'tls://', '', $parameters['host'] )
                 );
+                $args['ssl'] = $parameters['ssl'];
             }
```
tillkruss, line 714: The `WP_REDIS_TLS` constant needs to be shown in the diagnostics.

includes/object-cache.php
tillkruss, line 712: This doesn't work with PhpRedis. As a general rule, we don't merge changes that haven't been tested thoroughly, because it puts too many sites at risk.

tillkruss (changes requested):

includes/object-cache.php
david-donachie: I already responded to this. As far as I can see it *is* shown in the diagnostics, because you do `$this->diagnostics += $args;` on line 743.

Am I misunderstanding?

tillkruss: Yes, the constant itself needs to be listed as well.

david-donachie: Ahh! includes/diagnostics.php:68?

david-donachie:
I believe I've done all the requested changes

* WP_REDIS_SSL changed to WP_REDIS_TLS
* WP_REDIS_TLS added to README.md and to diagnostics.php
* WP_REDIS_TLS parsed to $parameters['ssl'] for predis
* WP_REDIS_TLS parsed to $args['others']['stream'] for phpredis > 5.3.0
* Tested with both predis and phpredis

david-donachie:
@tillkruss I *think* I've made all the changes you requested, but I still see the PR marked as requesting further changes. Is there something I've not yet done? Our deadline is approaching quickly, so if you need me to do something else, just let me know.

includes/object-cache.php
```diff
@@ -640,7 +640,7 @@ protected function build_parameters() {
             'database',
             'timeout',
             'read_timeout',
-            'retry_interval',
+            'retry_interval'
```
tillkruss, line 643: This comma can stay.

README.md
```diff
@@ -97,6 +98,15 @@ define( 'WP_REDIS_HOST', 'master.ncit.ameaqx.use1.cache.amazonaws.com' );
...
 
+Additional stream connection options for `phpredis` and `predis` connections can be defined using `WP_REDIS_TLS`
+
+```php
+define( 'WP_REDIS_TLS', [
+    'verify_host' => false,
+    'verify_host_name' => false
+]);
+```
+
```
tillkruss, line 109: Now seeing the code, I think we need to change the name. What this constant really is are context options: https://www.php.net/manual/en/context.ssl.php

Let's call it `WP_REDIS_SSL_CONTEXT` to stick with the PHP naming.

README.md
```diff
@@ -60,6 +60,7 @@ The plugin comes with vast set of configuration options.
 | `WP_REDIS_DISABLE_METRICS`           | `false`     | Disables metrics collection and display |
 | `WP_REDIS_DISABLE_BANNERS`           | `false`     | Disables promotional banners |
 | `WP_REDIS_DISABLE_DROPIN_AUTOUPDATE` | `false`     | Disables the drop-in auto-update |
+| `WP_REDIS_TLS`                       |             | TLS connection options for `tls` or `rediss` scheme |
```
tillkruss, line 63: Let's make the default an empty array `[]` and link to https://www.php.net/manual/en/context.ssl.php in the description.

includes/object-cache.php
```diff
@@ -702,23 +702,26 @@ protected function connect_using_phpredis( $parameters ) {
...
+
             if ( strcasecmp( 'tls', $parameters['scheme'] ) === 0 ) {
                 $args['host'] = sprintf(
                     '%s://%s',
                     $parameters['scheme'],
                     str_replace( 'tls://', '', $parameters['host'] )
                 );
+                if ( version_compare( $version, '5.3.0', '>=' ) && defined( 'WP_REDIS_TLS' ) && !empty(WP_REDIS_TLS) ) {
+                    $args['others']['stream'] = WP_REDIS_TLS;
+                }
```
tillkruss, line 717: This looks good. You can do the same code for `connect_using_relay()`, the API is identical.

tillkruss (changes requested):

README.md
david-donachie: Changed

includes/object-cache.php
david-donachie: At what version of Relay does the stream option become supported? I can see that you maintain the package. Is ir [0.6.2](https://github.com/cachewerk/relay/releases/tag/v0.6.2) where you mention changing the arity of connect()?

README.md
david-donachie: Done

includes/object-cache.php
tillkruss: You can skip the version check for Relay. Most of them support it.

PR #454, opened 2023-04-18
Use correct placeholders for admin bar title

ocean90:
Using `%d` for strings makes the title having only rounded numbers:

Bildschirm­foto 2023-04-18 um 16 04 42

This also simplifies the translators comment per https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#descriptions


tillkruss:
Great catch, thank you @ocean90.

PR #456, opened 2023-04-21
Allow translation of Object Cache Pro headline

MatzeKitt:
Currently, the headline of the Pro advertising is not translatable. This pull request fixes this and allow its translation.

### DIFF
--- a/includes/ui/settings.php
+++ b/includes/ui/settings.php
@@ -74,8 +74,11 @@ defined( '\\ABSPATH' ) || exit;
 
                 <div class="card">
                     <h2 class="title" style="line-height: 1.4">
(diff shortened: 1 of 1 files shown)


tillkruss:
I'm not sure what the rules around premium stuff is in the plugin guidelines.

MatzeKitt:
I think we already had this discussion when I requested the translatability of the bullet points some time ago. 😄

The only thing I could find was regarding general SaaS: https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/#6-software-as-a-service-is-permitted

However, it’s explicitly allowed to use that. And you should explicitly allow to translate any content shipped within your plugin.

PR #466, opened 2023-05-01
Don't redirect to settings on flushing Redis cache

yatsukhnenko:
- Added new action to flush cache with ajax request


tillkruss:
I took your code and made a branch, because I was unable to commit to your `yatsukhnenko:issue-459`.

PR #467, opened 2023-05-01
Don't redirect to settings on flushing Redis cache

### LINKED ISSUE #459: Flushing Redis cache should not redirect to Redis settings
Currently, if you flush the cache in the admin bar from whatever page, you will be redirected to the settings page of the Redis cache. At least on multisites. I think this is a bad user experience, since there is no indication that you will be redirected to said page nor can it be argued as intended, since there is a dedicated settings link below.

tillkruss:
Replaces #466. Closes #459.

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -108,8 +108,8 @@ class Plugin {
         add_action( 'wp_dashboard_setup', [ $this, 'setup_dashboard_widget' ] );
         add_action( 'wp_network_dashboard_setup', [ $this, 'setup_dashboard_widget' ] );
 
-        add_action( 'wp_ajax_roc_dismiss_notice', [ $this, 'dismiss_notice' ] );
-        add_action( 'wp_ajax_flush_cache', [ $this, 'flush_cache' ] );
+        add_action( 'wp_ajax_roc_dismiss_notice', [ $this, 'ajax_dismiss_notice' ] );
+        add_action( 'wp_ajax_roc_flush_cache', [ $this, 'ajax_flush_cache' ] );
 
         add_filter( 'plugin_row_meta', [ $this, 'add_plugin_row_meta' ], 10, 2 );
         add_filter( sprintf( '%splugin_action_links_%s', is_multisite() ? 'network_admin_' : '', WP_REDIS_BASENAME ), [ $this, 'add_plugin_actions_links' ] );
@@ -690,7 +690,7 @@ class Plugin {
 
                         try {
                             var data = new FormData();
-                            data.append('action', 'flush_cache');
+                            data.append('action', 'roc_flush_cache');
                             data.append('nonce', '{$nonce}');
 
                             var response = await fetch('{$ajaxurl}', { method: 'POST', body: data });
@@ -744,7 +744,7 @@ HTML;
         );
 
         $value = $this->get_status();
-        $title = __( 'Status: ', 'redis-cache' ) . $value;
+        $title = sprintf( __( 'Status: %s', 'redis-cache' ), $value );
 
         if ( $redis_status ) {
             global $wp_object_cache;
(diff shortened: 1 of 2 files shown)


includes/class-plugin.php
```diff
@@ -665,38 +666,68 @@ public function render_admin_bar( $wp_admin_bar ) {
...
+
+                            var responseText = await response.text();
+                            textNode.innerText = responseText;
+
+                            setTimeout(function () {
+                                textNode.innerText = 'Object Cache';
+                            }, 3000);
+                        } catch (error) {
+                            textNode.innerText = responseText;
+                            alert('Object cache could not be flushed: ' + error);
```
tillkruss, line 699: This needs to be tested.

includes/class-plugin.php
```diff
@@ -665,38 +666,68 @@ public function render_admin_bar( $wp_admin_bar ) {
...
         }
 
+        $html = <<<HTML
+            <style>
+                #wpadminbar ul li.redis-cache-error { background: #c00; }
+                #wpadminbar:not(.mobile) .ab-top-menu > li.redis-cache-error:hover > .ab-item { background: #b30000; color: #fff; }
+            </style>
+            <script>
+                document.querySelector('#wp-admin-bar-redis-cache-flush > a')
+                    .addEventListener('click', async function (event) {
```
tillkruss, line 676: We have to use Vanilla js, because the `admin.js` wasn't loaded on the frontend.

includes/class-plugin.php
```diff
@@ -665,38 +666,68 @@ public function render_admin_bar( $wp_admin_bar ) {
...
+                        var node = document.querySelector('#wp-admin-bar-redis-cache');
+                        var textNode = document.querySelector('#wp-admin-bar-redis-cache > .ab-item:first-child');
+
+                        node.classList.remove('hover');
+                        textNode.innerText = 'Flushing cache...';
+
+                        try {
+                            var response = await fetch(qm_l10n.ajaxurl, {
+                                method: 'POST',
+                                body: 'action=flush_cache&nonce=...'
```
tillkruss, line 688: We need to embed a `nonce` here, ideally using PHP: https://developer.wordpress.org/reference/functions/wp_create_nonce/

includes/class-plugin.php
```diff
@@ -665,38 +666,68 @@ public function render_admin_bar( $wp_admin_bar ) {
...
+            <script>
+                document.querySelector('#wp-admin-bar-redis-cache-flush > a')
+                    .addEventListener('click', async function (event) {
+                        event.preventDefault();
+
+                        var node = document.querySelector('#wp-admin-bar-redis-cache');
+                        var textNode = document.querySelector('#wp-admin-bar-redis-cache > .ab-item:first-child');
+
+                        node.classList.remove('hover');
+                        textNode.innerText = 'Flushing cache...';
```
tillkruss, line 683: Can we make this a PHP variable, so it's translateable?

includes/class-plugin.php
yatsukhnenko: I'm wondering is it fine in WP mixing PHP and JS code? Maybe it would be better to use something like `wp_enqueue_script`?

includes/class-plugin.php
```diff
@@ -664,39 +665,73 @@ public function render_admin_bar( $wp_admin_bar ) {
...
+                            data.append('nonce', '{$nonce}');
+
+                            var response = await fetch('{$ajaxurl}', { method: 'POST', body: data });
+
+                            textNode.innerText = await response.text();
+
+                            setTimeout(function () { textNode.innerText = '{$nodeTitle}'; }, 3000);
+                        } catch (error) {
+                            textNode.innerText = '{$nodeTitle}';
+                            alert('Object cache could not be flushed: ' + error);
```
tillkruss, line 703: @yatsukhnenko Does this work?

PR #471, opened 2023-05-19
Syntax error occurred in the statement to obtain the password

xiaoman1221:
ERROR:Warning: strlen() expects parameter 1 to be string, array given inredis-cache/dependencies/predis/predis/src/Connection/Factory.php on line 167


tillkruss:
You cannot patch dependencies.

tillkruss:
See 0a6e486

PR #478, opened 2023-06-15
Update banner visibility rules for AccelerateWP installs.

martinkrcho:
Changes the logic for displaying Object Cache PRO banners on the settings page and in the plugins' list. Constant `WP_REDIS_DISABLE_BANNERS` now affects visibilities of these banners only for AccelerateWP installs.

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -279,9 +279,13 @@ class Plugin {
             return $plugin_meta;
         }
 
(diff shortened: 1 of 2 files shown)


tillkruss:
@martinkrcho: I think we can release this now, see #478. Any more thoughts?

martinkrcho:
I left a [comment](https://github.com/rhubarbgroup/redis-cache/commit/1f12771b2a83660da203cf79fe2df9a238f60d89#r119257455) on your commit.

PR #479, opened 2023-06-23
Refactor AccelerateWP installation check.

martinkrcho:
Contains a refactor of `Plugin::acceleratewp_install()` to take the `WP_REDIS_DISABLE_BANNERS` constant into account.


tillkruss:
@martinkrcho: I think we can release this now ✌️

martinkrcho:
I agree, go for it 👍🏼

PR #486, opened 2023-07-20
Handle unexpected result from Redis pipeline quietly.

martinkrcho:
This pull requests changes the behaviour of caching functions that utilize Redis pipeline to execute multiple commands in batch. When the pipeline returns unexpected results, it no longer logs an error, but discard the current transaction and returns an empty array.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -1300,7 +1300,7 @@ class WP_Object_Cache {
 
             if ( count( $results ) !== count( $keys ) ) {
                 $tx->discard();
-                return [];
+                return array_fill_keys( $keys, false );
             }
 
             $results = array_combine( $keys, $results );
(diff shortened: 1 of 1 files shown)


includes/object-cache.php
```diff
@@ -1299,7 +1299,8 @@ protected function add_multiple_at_once( array $data, $group = 'default', $expir
             }, $tx->{$method}() ?: [] );
 
             if ( count( $results ) !== count( $keys ) ) {
-                throw new Exception( 'Redis pipeline returned unexpected result' );
+                $tx->discard();
+                return [];
```
tillkruss, line 1303: WP's API is a bit stupid, but I think these should stick with the expected return value:

```php
array_fill_keys($keys, false);
```

tillkruss (changes requested):

includes/object-cache.php
martinkrcho: Of course. Updated :)

PR #487, opened 2023-08-12
Fix typos

szepeviktor:
Found two misspellings.


szepeviktor:
More goodies
https://github.com/szepeviktor/byte-level-care/blob/61ca4439630a47d4e6946173a369270a4729bfb2/.github/workflows/spelling.yml#L31-L50
🎈 more fun.

PR #494, opened 2023-10-25
Check if WP_REDIS_PLUGIN_PATH is defined before defining

rbackert:
When using this as a must-use plugin, developers must define `WP_REDIS_PLUGIN_PATH` so the drop-in can locate files. The main plugin file also defines this constant, but without checking if it's already been defined, triggering PHP warnings.

Check if constant is defined before defining it, to avoid warnings.


tillkruss:
Thanks, can you add a CHANGELOG entry too?

PR #496, opened 2023-11-17
Fix AccelerateWP installation check

sy-records:
The sidebar in the settings page is now not controlled by `WP_REDIS_DISABLE_BANNERS`, which should be wrong?

https://github.com/rhubarbgroup/redis-cache/blob/1f12771b2a83660da203cf79fe2df9a238f60d89/includes/ui/settings.php#L68-L72

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -1611,6 +1611,10 @@ HTML;
             }
         }
 
-        return false;
+        if ( $ignore_banner_constant ) {
+            return false;
+        }
+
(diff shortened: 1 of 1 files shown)


tillkruss:
Can you explain what's currently happening, and what you expect to happen?

sy-records:
I want to hide the pro ad sidebar in the settings page. 

Up to now, WP_REDIS_DISABLE_BANNERS does not take effect and cannot be hidden.

PR #504, opened 2024-02-29
Actually test for ext-relay while running PHPStan

szepeviktor:
@tillkruss The comment made me do this.


szepeviktor:
Surprise!

shivammathur/setup-php has no relay extension for PHP 7.2

PR #518, opened 2024-04-23
adds support for ssl context in redis clusters

mbrowngold:
This is necessary when connecting to a REDIS cluster that requires an ssl context (e.g. AWS Elasticache serverless)

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -683,7 +683,7 @@ class WP_Object_Cache {
                     'cluster' => $this->build_cluster_connection_array(),
                     'timeout' => $parameters['timeout'],
                     'read_timeout' => $parameters['read_timeout'],
-                    'persistent' => $parameters['persistent']
+                    'persistent' => $parameters['persistent'],
                 ];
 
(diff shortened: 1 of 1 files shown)


includes/object-cache.php
```diff
@@ -684,11 +684,17 @@ protected function connect_using_phpredis( $parameters ) {
                     'timeout' => $parameters['timeout'],
                     'read_timeout' => $parameters['read_timeout'],
                     'persistent' => $parameters['persistent'],
+                    'password' => null,
```
tillkruss, line 687: We might not be able to just set it and have to set it to null or the actually password in the next condition to not pass to many parameters.

tillkruss (changes requested):

includes/object-cache.php
mbrowngold: I've modified the PR so that the password key is only set when it is known that version is sufficiently high.

tillkruss:
Great. Can you test the changes I've made. If they work for you we can merge.

mbrowngold:
> Great. Can you test the changes I've made. If they work for you we can merge.

it works, but I believe you meant to do `!array_key_exists()` (I don't have a password on my setup so it would still work for me)

PR #521, opened 2024-05-03
add test for flush group

Trisky:


Trisky:
Tests seem to pass. It seems that sentinel is not being tested.

tillkruss:
@Trisky: No, feel free to add unit tests. 

Object Cache Pro has over 1500 test cases, if you need something reliable. Happy to send a discount code.

PR #522, opened 2024-05-06
Fix Redis version determination for predis replication connection

yatsukhnenko:
The first commit just suppress warning when Redis server version couldn't be determined with `INFO` command.
In the second commit I tried to get `INFO` results from the master node for `MasterSlaveReplication` connection


yatsukhnenko:
image

PR #535, opened 2024-06-06
Add filter to allow customizing who can flush the cache

pmgarman:
Couldn't find a CONTRIBUTING.md (i may be blind) so took a best guess. Also just assumed 2.6 but who knows?!

I think this could easily be a thing that gets enhanced deeper for multisite to allow individual site owners to flush their individual sites cache or something along those lines... but this probably serves my needs well enough :D

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 - Fixed several issues with Predis and cluster/replicated connection
 - Fixed rare fatal error in `show_error_and_die()` (one more time)
-- Add `redis_cache_manager_capability` filter to allow customizing who can manage the plugin
-- Add `WP_REDIS_MANAGER_CAPABILITY` constant to allow for short circuiting logic to determine which capability to check for users who can manage the plugin
+- Add `redis_cache_manager_capability` filter
+- Add `WP_REDIS_MANAGER_CAPABILITY` constant
 
 ## 2.5.2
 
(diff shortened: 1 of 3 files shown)


pmgarman:
This was the code i used to test that this worked for the author role:
```php
add_filter( 'roc_manage_redis_capability', function( $cap ) {
	return 'publish_posts';
});
```

pmgarman:
Resolved conflict in changelog

tillkruss:
Should we also make this a constant for easier access?

pmgarman:
You would only be able to define your default as a constant though right? Since the constant can't change after being filtered?

tillkruss:
> You would only be able to define your default as a constant though right? Since the constant can't change after being filtered?

Yeah, it would be set once in the `wp-config.php` and should be opt-in, not used by default.

pmgarman:
Oh I see what you mean, inverse of what I was thinking. It's monday...

I could see that being handy too yeah. I can update my PR and fix those codestyle complaints.

Any preference on your constant name?

pmgarman:
Nevermind those errors are yours :P enjoy!

pmgarman:
Opted for `WP_REDIS_MANAGER_CAPABILITY` until @tillkruss says otherwise :)

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Added `roc_manage_redis_capability` filter and `WP_REDIS_MANAGER_CAPABILITY` constant
```
tillkruss, line 5: @pmgarman: Could you add these to the README as well?

tillkruss (changes requested):

pmgarman:
@tillkruss try this on for size

PR #546, opened 2024-09-06
Respect WP_REDIS_SCHEME for cluster connection

yatsukhnenko:
Fixes https://github.com/predis/predis/issues/1474

### DIFF
--- a/includes/class-predis.php
+++ b/includes/class-predis.php
@@ -86,6 +86,11 @@ class Predis {
             $servers = $this->build_cluster_connection_array();
             $parameters['cluster'] = $servers;
             $options['cluster'] = 'redis';
+            if ( defined('WP_REDIS_SCHEME') ) {
+                $options['parameters'] = [
+                    'scheme' => WP_REDIS_SCHEME,
+                ];
+            }
         }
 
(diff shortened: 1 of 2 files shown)


includes/class-predis.php
```diff
@@ -86,6 +86,11 @@ public function connect( $read_timeout = null ) {
             $servers = $this->build_cluster_connection_array();
```
tillkruss, line 86: Should we maybe do the scheme as part of `build_cluster_connection_array()`?

includes/class-predis.php
```diff
@@ -206,13 +206,30 @@ protected function build_cluster_connection_array() {
...
+                    $components['host'],
+                    $components['port']
+                );
+            } else {
+                $cluster[ $key ] = sprintf(
+                    '%s:%d',
+                    $components['host'],
+                    $components['port']
+                );
+            }
```
tillkruss, line 232: Cool! Do we need to copy this over to the same function in `object-cache.php`?

tillkruss (changes requested):

tillkruss (approved):

PR #548, opened 2024-09-09
Check for WP_REDIS_DISABLED environment variable

SpiZeak:
I am using a .env file in my projects and I would like WP_REDIS_DISABLED=true to disable Redis.


tillkruss:
You can use:

```php
define('WP_REDIS_DISABLED', getenv('WP_REDIS_DISABLED') ?: false);
```

PR #559, opened 2024-12-11
Invent cache salt

edneville:
If there are multiple sites using the same redis server then invent a salt that gives uniqueness for each site to avoid collision.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -249,6 +249,14 @@ function wp_cache_init() {
         define( 'WP_REDIS_PREFIX', getenv( 'HTTP_X_APP_USER' ) );
     }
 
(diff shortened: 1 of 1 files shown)


tillkruss:
This will suddenly give all sites a prefix and jack up their Redis server size. Feel free to extend the documentation, (or admin gui) instead.

edneville:
@tillkruss, thanks for reviewing. Isn't it better to have a prefix to avoid collision? Would it be acceptable to then to say in the docs to define WP_REDIS_PREFIX with an empty string if that's not desired? Ideally I'd like to provide some default protection to avoid two installs on the same infrastructure from colliding.

tillkruss:
Yeah, feel free to document it, or even add it to the admin UI.

PR #562, opened 2025-01-01
Move common code to AbstractReplication class

yatsukhnenko:


tillkruss (commented):
We pull Predis from GitHub via Composer, so any update would nuke this. Can we find a different solution?

PR #566, opened 2025-01-24
Suppress Squiz.PHP.Heredoc.NotAllowed warning

yatsukhnenko:
Disable `Squiz.PHP.Heredoc.NotAllowed` before heredoc

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -827,6 +827,7 @@ class Plugin {
      * @return string
      */
     protected function admin_bar_style() {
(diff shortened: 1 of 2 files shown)


tillkruss:
How can this be disabled globally, without using comments?

yatsukhnenko:
If I correctly understand to disable check globally it requires changes in `phpcs.xml.dist` but I'm not sure how to do this in GH actions. I can try to add CI step to find and modify config before running checks

PR #587, opened 2025-09-15
Updated wp_cache_*() fn signatures to match core

### LINKED ISSUE #586: Function parameter mismatch breaks named arguments
## Description
WordPress core cache functions use the parameter name `$expire`, like in `function wp_cache_add( $key, $data, $group = '', $expire = 0 )` but the `object-cache.php` created by this plugin uses `$expiration` instead, `function wp_cache_set( $key, $value, $group = '', $expiration = 0 )`.  This mismatch wasn't an issue until PHP introduced named arguments. I just discovered the mismatch when I was testing code with this plugin enabled and my cache call failed with a fatal error.

## Expected Behavior
A function call like `wp_cache_set( $cache_key, $results, expire: 3600);` should succeed whether or not the Redis Object Cache plugin is enabled.

## Actual Behavior
With the plugin enabled, it fails with the error `PHP Fatal error:  Uncaught Error: Unknown named parameter $expire`

## Possible Fix
Rename the parameter from `$expiration` to `$expire` so it matches WordPress's function signature or accept both for compatibility.

## Environment

* Plugin version: 2.6.5
* PHP version: 8.2.29
* WordPress version: 6.8.2

Enchiridion:
This fixes #586 by updating the wp_cache_* function signatures to match core.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -50,17 +50,17 @@ function wp_cache_supports( $feature ) {
  * If the specified key already exists, the value is not stored and the function
  * returns false.
  *
- * @param string $key        The key under which to store the value.
- * @param mixed  $value      The value to store.
- * @param string $group      The group value appended to the $key.
- * @param int    $expiration The expiration time, defaults to 0.
+ * @param string $key    The key under which to store the value.
+ * @param mixed  $data   The value to store.
+ * @param string $group  The group value appended to the $key.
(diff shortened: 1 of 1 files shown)


tillkruss:
Thanks! Looks like Object Cache Pro is already matching core. I won't release this right away but feel free to pull `dev-develop` if you want to use it right away.

Issue #574, opened 2025-07-17
`igbinary` issue with `wp_cache_incr()`


colart-griosa:
## Description
A recent update to WooCommerce which utilizes `wp_cache_incr` was giving us Redis errors when creating new orders. We followed the root of the issue and we noticed that because we were using the `WP_REDIS_IGBINARY = true` function, the cached value was being stored in binary format. 

However when the `wp_cache_incr` function is run it expects the cached value to be an integer, giving an error. I [created a ticket on the WooCommerce repo](https://github.com/woocommerce/woocommerce/issues/59029) initially, which lead me to find the root problem.

## Expected Behavior
`wp_cache_incr` should be able to increment a key stored in binary format when `WP_REDIS_IGBINARY = true`

## Actual Behavior
I get the following error
```
Predis\Response\ServerException: ERR value is not an integer or out of range in /var/www/html/wp-content/plugins/redis-cache/dependencies/predis/predis/src/Client.php:384
Stack trace:
#0 /var/www/html/wp-content/plugins/redis-cache/dependencies/predis/predis/src/Client.php(353): Predis\Client->onErrorResponse()
#1 /var/www/html/wp-content/plugins/redis-cache/dependencies/predis/predis/src/Client.php(303): Predis\Client->executeCommand()
#2 /var/www/html/wp-content/object-cache.php(2412): Predis\Client->__call()
#3 /var/www/html/wp-content/object-cache.php(223): WP_Object_Cache->increment()
#4 /var/www/html/wp-content/plugins/woocommerce/src/Caches/OrderCountCache.php(134): wp_cache_incr()
#5 /var/www/html/wp-content/plugins/woocommerce/src/Caches/OrderCountCacheService.php(104): Automattic\WooCommerce\Caches\OrderCountCache->increment()
#6 /var/www/html/wp-includes/class-wp-hook.php(324): Automattic\WooCommerce\Caches\OrderCountCacheService->update_on_new_order()
#7 /var/www/html/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters()
#8 /var/www/html/wp-includes/plugin.php(517): WP_Hook->do_action()
#9 /var/www/html/wp-content/plugins/woocommerce/includes/data-stores/class-wc-order-data-store-cpt.php(113): do_action()
#10 /var/www/html/wp-content/plugins/woocommerce/includes/class-wc-data-store.php(186): WC_Order_Data_Store_CPT->create()
[...]

PR #588, opened 2025-09-26
Revert incr/decr changes

tillkruss:
We need to keep the TTL of keys when `wp_cache_(in|de)cr()` is called.

1. Do we drop the `igbinary_(un)serialize()` feature all together? This keeps the `INCR` to a single command.
2. Do we read the TTL (and set it) when using igbinary? This means 3 commands are fired for each INCR if igbinary is enabled.

- - - 

### Alternatives
- Do we require Redis 6.x and use `KEEPTTL`? (health check to promote risk and upgrade)
- Do we limit igbinary to PhpRedis 6.2.0 and use `Redis::OPT_PACK_IGNORE_NUMBERS`?

See #574.

### DIFF
--- a/includes/object-cache.php
+++ b/includes/object-cache.php
@@ -519,18 +519,11 @@ class WP_Object_Cache {
 
         $this->cache_group_types();
 
-        if ( defined( 'WP_REDIS_TRACE' ) && WP_REDIS_TRACE ) {
-            trigger_error('Tracing feature was removed', E_USER_DEPRECATED);
-        }
-
         $client = $this->determine_client();
         $parameters = $this->build_parameters();
 
         try {
             switch ( $client ) {
-                case 'hhvm':
-                    $this->connect_using_hhvm( $parameters );
-                    break;
                 case 'phpredis':
                     $this->connect_using_phpredis( $parameters );
                     break;
@@ -600,7 +593,7 @@ class WP_Object_Cache {
         $client = 'predis';
 
         if ( class_exists( 'Redis' ) ) {
-            $client = defined( 'HHVM_VERSION' ) ? 'hhvm' : 'phpredis';
+            $client = 'phpredis';
         }
 
         if ( defined( 'WP_REDIS_CLIENT' ) ) {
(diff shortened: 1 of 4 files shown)


includes/object-cache.php
```diff
@@ -2410,13 +2351,18 @@ public function increment( $key, $offset = 1, $group = 'default' ) {
         }
 
         try {
-            $value = (int) $this->parse_redis_response( $this->maybe_unserialize( $this->redis->get( $derived_key ) ) );
-            $value += $offset;
-            $result = $this->parse_redis_response( $this->redis->set( $derived_key, $this->maybe_serialize( $value ) ) );
+            if ( $this->use_igbinary ) {
+                $value = (int) $this->parse_redis_response( $this->maybe_unserialize( $this->redis->get( $derived_key ) ) );
+                $value += $offset;
+                $result = $this->parse_redis_response( $this->redis->set( $derived_key, $this->maybe_serialize( $value ) ) );
```
tillkruss: @yatsukhnenko Can you add a call for `PTTL` and set it with `SET PX` if the key had an expiry? For `increment()` and `decrement()`.

Probably needs to be tested against Predis and PhpRedis, not sure if they take the same arguments in `set()`

includes/object-cache.php
```diff
@@ -2342,13 +2351,24 @@ public function increment( $key, $offset = 1, $group = 'default' ) {
...
-            $value = (int) $this->parse_redis_response( $this->maybe_unserialize( $this->redis->get( $derived_key ) ) );
-            $value += $offset;
-            $result = $this->parse_redis_response( $this->redis->set( $derived_key, $this->maybe_serialize( $value ) ) );
+            if ( $this->use_igbinary ) {
+                $value = (int) $this->parse_redis_response( $this->maybe_unserialize( $this->redis->get( $derived_key ) ) );
+                $value += $offset;
+                $serialized = $this->maybe_serialize( $value );
+
+                if ( ($pttl = $this->redis->pttl( $derived_key )) > 0 ) {
+                    $result = $this->parse_redis_response( $this->redis->set( $derived_key, $serialized, [ 'px' => $pttl ] ) );
```
tillkruss: Predis won't take an array.

yatsukhnenko (approved):

tillkruss:
Merged via https://github.com/rhubarbgroup/redis-cache/commit/d245255b5d8cbc54d0153b6d6af931e0a8284791.

tillkruss:
@YachYaroslav v2.7.0 was release just now.

PR #589, opened 2025-10-06
Delete transients when activating cache

tillkruss:
This should avoid stale WooCommerce data.

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -1210,10 +1210,10 @@ HTML;
      * Delete all transients if the cache was enabled successfully.
      * Callback for `redis_object_cache_enable` action.
      *
-     * @param bool $result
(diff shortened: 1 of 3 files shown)


includes/class-plugin.php
```diff
@@ -1204,6 +1206,33 @@ public function register_shutdown_hooks() {
...
     }
 
+    /**
+     * Delete all transients if the cache was enabled successfully.
+     * Callback for `redis_object_cache_enable` action.
+     *
+     * @param bool $result
+     * @return void
+     */
+    public function maybe_delete_transients( $result ) {
```
yatsukhnenko: I think there should be some meaningful variable and not just `$result`

yatsukhnenko (approved):

PR #595, opened 2025-11-27
Fix - Correct ABSPATH constant check

shivapoudel:
This PR fixes the `ABSPATH` constant validation. The previous `\\ABSPATH` check was invalid and caused scripts to exit. It now uses the correct `ABSPATH` check, following the standard WordPress practice for secure file access.


tillkruss:
I want to say the correct syntax would be \ABSPATH so that it always checks the root namespace. Can you confirm that still fixes it for you?

PR #603, opened 2026-01-30
Update apexcharts.js to the latest stable version

EusebiuOprinoiu:
This commit updates the `apexcharts.js` library to the latest stable version.
It resolves conflicts with other plugins that use the same library, as described in [this thread](https://wordpress.org/support/topic/conflict-between-merchant-and-redis-object-cache/) on the WordPress Forum.

The jump from version 3.x to version 5.x comes with very minor CSS changes, the most notable one being that Markers are now squares (instead of circles) and that the labels have a larger font size.

The CSS tweaks in the `admin.css` file address these changes to ensure the widget design remains unchanged.


EusebiuOprinoiu:
I'm not sure how external libraries should be handled when multiple plugins bundle them, so I leave that up to you if you want minimize conflicts. Ideally, a library should only be loaded once.
At least with the latest version, the widget from the admin dashboard no longer breaks.

I also notice that with the old version, the highest peak of a chart used to reach the top of the graph.
With the new version, the highest peak is lower. (at least for the Times, Bytes, and Calls tabs - for the Ratio tab, it still reaches the top)
I presume some minor changes need to be made in the `admin.js` file as well, but I'm not very good with JavaScript, so I leave that up to you.

tillkruss:
Redis Object Cache loads the charts on these screens: `roc-settings, dashboard, dashboard-network`. Can you confirm where your other plugin loads the charts?

EusebiuOprinoiu:
The other plugin loads the charts only on the main dashboard. ( `dashboard` )
If you need to test, you can install this plugin:
https://wordpress.org/plugins/merchant

EusebiuOprinoiu:
Hello, Till!
Any updates about this?

tillkruss:
@EusebiuOprinoiu I forgot about this, dang.

~I can get this merged today, but if this will cause compatibility issues with other plugins I'll revert it.~

assets/css/admin.css
```diff
@@ -158,7 +158,8 @@
 }
 
 #redis-stats-chart .apexcharts-tooltip,
-#redis-stats-chart .apexcharts-tooltip-title {
+#redis-stats-chart .apexcharts-tooltip-title,
+#redis-stats-chart .apexcharts-tooltip-marker {
     font-size: 13px;
```
tillkruss, line 163: Looking at the search results, this is much trickier than expected: https://veloria.dev/search/48597b46-43e0-4e04-b9d7-9f9b15518bf3

What do you think about adding a toggle to set the apexchart version, 3-5 so you have control over it?

tillkruss (changes requested):

assets/css/admin.css
EusebiuOprinoiu: I don't think there is a need for a version switcher, but I'm not opposed to it.
Things break only for plugins that use version 3 or older. If you update the library to version 5, you eliminate almost all potential conflicts.

tillkruss: If 5.x is backwards compatible then cool. I just saw plenty of other plugins use version 3 and 4.

EusebiuOprinoiu: From my tests, versions 4 and 5 are compatible. Only version 3 is problematic.
If other plugins bundle very old versions and cause issues, it's up to them to provide a fix.
And because conflicts occur only when those plugins load their libraries in the main admin dashboard, the chances of a conflict drop significantly.

PR #604, opened 2026-03-10
Fix Predis Sentinel authentication when Sentinel requires password

bsabalaskey:
Problem: WP_REDIS_PASSWORD is applied to Redis node parameters but not Sentinel-specific parameters in Predis Sentinel mode.

Symptom: admin health check shows Redis as unreachable unless each Sentinel URI includes ?password=whatever

Fix: propagate Redis password and optional username into $servers when WP_REDIS_SENTINEL is defined.

Why it matters: avoids forcing credentials into each Sentinel URI and makes authenticated Sentinel setups work consistently and easier.

### Reproduce the Problem and Showing the Fix

1. Configure authenticated Redis Sentinel
2. Set WP_REDIS_CLIENT = 'predis'. 
3. Set WP_REDIS_SENTINEL and WP_REDIS_SERVERS (like 'tcp://1.2.3.4:26379')
4. Set WP_REDIS_PASSWORD
5. Observe NOAUTH Authentication required; plugin reports Redis unreachable and not possible to enable object caching
6. Apply changes from this code
7. Connection succeeds and you can enable object cache

### Tested With

- Redis Sentinel 3-node setup (separate from WordPress servers), Redis v8.6.1
- Redis and Sentinel password set to identical values, with default user (no ACL)
- Predis client v2.4.0
- Authenticated Sentinel and Redis using same password (they must be identical)
- Confirmed admin check and runtime connection both work after patch 
- PHP Version: 8.2.30 with NGINX using PHP FPM
- WordPress v6.9.1
- Default Redis port of 6379 and Sentinel port 26379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 - Delete transients when cache is enabled
 
+# 2.7.1
+
+- Support password configuration for Sentinel connections when using Predis (from @bsabalaskey)
+- Configuration uses `WP_REDIS_SENTINEL_PASSWORD` constant
+
 ## 2.7.0
 
 - Preserve key TTL when calling (in|de)crement methods
--- a/README.md
+++ b/README.md
@@ -47,24 +47,25 @@ The Redis Object Cache plugin comes with vast set of configuration options. If y
 <details>
 <summary>Advanced configuration options</summary>
 
-| Configuration constant               | Default     | Description                                   |
-| ------------------------------------ | ----------- | --------------------------------------------- |
-| `WP_CACHE_KEY_SALT`                  |             | Deprecated. Replaced by `WP_REDIS_PREFIX` |
-| `WP_REDIS_FLUSH_TIMEOUT`             | `5`         | Experimental. The timeout in seconds when flushing |
-| `WP_REDIS_RETRY_INTERVAL`            |             | The number of milliseconds between retries (PhpRedis only) |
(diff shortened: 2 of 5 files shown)


tillkruss:
Can you think of any scenarios where this would cause issues for any existing Sentinel setups that don't use your exact password setup?

bsabalaskey:
That is why I excluded changing things with the password in the Sentinel URI. If there is no password set at all, then this code is not triggered. And this only applies to Sentinel. So I believe that I covered the cases. The only limit is that the Redis and Sentinel password have to be the same. I suppose that flexibility could be added by having a WP_SENTINEL_PASSWORD to allow that being different, and then if same is not defined, fall back to the WP_REDIS_PASSWORD.

tillkruss:
>The only limit is that the Redis and Sentinel password have to be the same.

That seems reasonable. Thanks!

tillkruss (changes requested):
This just needs a changelog entry.

bsabalaskey:
To make it flexible for a different Sentinel password from the Redis password, I used a WP_REDIS_SENTINEL_PASSWORD constant. The configuration style of that constant work in the same way as WP_REDIS_PASSWORD. Docs are all updated to describe that too.

tillkruss (changes requested):
Please add a changelog entry to your branch, then I can merge this ✌️

bsabalaskey:
The change was updated in the latest push I did, as part of the flexibility to add use a Sentinel password different from the Redis password.

PR #340, opened 2022-02-19
Optimisations

colinleroy:
Hello,

I hope you'll consider this PR. It aims at shaving a bit of performance (about 20% in redis-cache on my homepage) with two modifications:
- Avoid walking full arrays to determine group type: We use an associative array to get the type of the group instead of walking an array.
- Avoid sanitizing key/group multiple times in the same codepath: for example, ->get() was using sanitize_key to build the key and to check if the group was global.

I'll attach screenshots of KCachegrind that demonstrate the changes.

Hope this helps,
Colin

PR #606, opened 2026-03-26
remove unnecessary group_types array and associated methods.

nickchomey:
the performance optimizations introduced with the associative array in https://github.com/rhubarbgroup/redis-cache/pull/340 didnt actually require adding the group_types array and associated methods. Better to just convert global/ignored/unflushable_groups themselves to associative arrays and use isset on those directly.

### DIFF
--- a/includes/diagnostics.php
+++ b/includes/diagnostics.php
@@ -121,37 +121,43 @@ if ( isset( $info['WP_REDIS_SERVERS'] ) ) {
 }
 
 if ( $dropin && ! $disabled ) {
+    // Normalize group arrays to a plain list of names regardless of whether stored as
+    // an associative set (group_name => true) or a legacy indexed list.
+    $normalize_group_names = static function ( $groups ) {
+        $groups = (array) $groups;
+        if ( empty( $groups ) ) {
+            return array();
+        }
+        $keys = array_keys( $groups );
+        return ( $keys === range( 0, count( $groups ) - 1 ) ) ? array_values( $groups ) : $keys;
+    };
+
     $info['Global Groups'] = wp_json_encode(
-        array_keys( (array) ( $wp_object_cache->global_groups ?? [] ) ),
+        $normalize_group_names( $wp_object_cache->global_groups ?? [] ),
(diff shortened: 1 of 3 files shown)


includes/diagnostics.php
```diff
@@ -10,153 +10,177 @@
...
+$disabled   = defined( 'WP_REDIS_DISABLED' ) && WP_REDIS_DISABLED;
 
 $info['Status'] = $roc->get_status();
 $info['Client'] = $roc->get_redis_client_name();
 
 $info['Drop-in'] = $roc->object_cache_dropin_exists()
-    ? ( $dropin ? 'Valid' : 'Invalid' )
-    : 'Not installed';
+	? ( $dropin ? 'Valid' : 'Invalid' )
+	: 'Not installed';
```
tillkruss: Please revert all the whitespace changes.

tillkruss (changes requested):

includes/diagnostics.php
nickchomey: ugh, didnt notice that that happened. I definitely always try to avoid such changes in PRs. It happens automatically when I press Ctrl+S in my IDE. I'll revert tomorrow

tillkruss: thanks, no rush of course ✌️

nickchomey: done

nickchomey: nevermind. not done. will finish soon

nickchomey:
i admittedly did not test the new test, nor whether the old ones passed - not sure how to actually run them. Do you have any CI tests that do it? Or are you able to run yourself easily?

nickchomey:
@tillkruss were you able to look at the tests for this?

tillkruss:
@nickchomey I was under the impression that we're waiting on something in SQL object cache to be figured out first, is that one merged?

nickchomey:
It hasn't been merged in sqlite object cache either. But it's not clear to me how that is relevant? I'm happy to address anything that remains of concern here

PR #618, opened 2026-05-05
Upgrade apexcharts to 5.x

tillkruss:
Resolves #603.


tillkruss:
@EusebiuOprinoiu Please review.

EusebiuOprinoiu:
It looks good!
Thank you very much.

EusebiuOprinoiu:
Actually, you should consider updating these CSS rules to preserve the same styles as the old version. (and to have the same design in the main dashboard and on the plugin page)

```
#redis-stats-chart .apexcharts-tooltip-marker {
    width: 12px;
    height: 12px;
    margin-right: 7px;
}

#dashboard_rediscache .apexcharts-tooltip-marker {
    width: 12px;
    height: 12px;
    margin-right: 7px;
}
```

Basically, set the width and height to 12px for both rules, and add an explicit margin-right of 7px.

PR #633, opened 2026-08-28
Support persistent connections via WP_REDIS_PERSISTENT

### LINKED ISSUE #632: Connection parameter `persistent` is built but never used, except on PhpRedis clusters
`WP_Object_Cache::build_parameters()` builds a `persistent` connection parameter, defaulting to
`false`, but nothing ever sets it and almost nothing reads it:

- `persistent` is absent from the `$settings` list right below it, so no `WP_REDIS_PERSISTENT`
  constant is ever read into it. It stays `false` for the lifetime of the request, on every
  installation.
- The only place it is read is the `RedisCluster` branch of `connect_using_phpredis()`, where it
  is therefore always `false`.
- `connect_using_predis()` hands `$parameters` straight to `new Predis\Client()`, and Predis does
  honour a `persistent` parameter — so on Predis this is a supported option that the plugin
  explicitly forces off.
- `connect_using_phpredis()` and `connect_using_relay()` call `connect()`, with the fourth
  argument hardcoded to `''`. That argument is `reserved` for `connect()` and `persistent_id` for
  `pconnect()`, so the slot the feature would need is already there, unused.
- `connect_using_credis()` passes `'persistent' => ''` to the `Credis_Client` constructor, whose
  fourth argument is exactly its persistence identifier.

The net effect is that every PHP request opens a new TCP connection and a new RESP handshake to
Redis and closes it at the end of the request, with no supported way to opt out.

## Expected Behavior

A site should be able to reuse its Redis connection across requests, as the parameter already
present in the source suggests was intended.

## Actual Behavior

There is no way to enable it. `total_connections_received` on the Redis server grows by one per
PHP request. Measured on WordPress 6.9 / PHP 8.5-FPM / PhpRedis 6.3.0 against plugin 2.8.0
unmodified: 30 requests that boot WordPress raise the counter by 31, the extra one being
background drift measured separately at rest.

## Possible Fix

Read a `WP_REDIS_PERSISTENT` constant into the parameter that already exists, and use it:
`pconnect()` instead of `connect()` on PhpRedis and Relay, and a non-empty identifier in the
`Credis_Client` constructor. Predis needs nothing beyond the constant.

One thing is not merely cosmetic: persistent connections are pooled per host, port and
identifier, so anything applied *after* connecting belongs in that identifier. `select()` is only
called when the database is non-zero, so two sites sharing a pool could otherwise inherit
whichever database the other one selected last. An identifier derived from the database (with an
override constant) avoids that.

A pull request implementing this is attached.

## Steps to Reproduce

1. Install the plugin against a Redis server, on PhpRedis, Relay or Credis.
2. Read `total_connections_received` from `INFO stats`.
3. Issue N page requests that boot WordPress.
4. Read it again: it has grown by roughly N, and no configuration constant changes that.

## Additional context

The cost is small per connection and entirely real under concurrency. Measured with
`memtier_benchmark` against a Redis-protocol server on loopback, 100 connections, isolated from
any WordPress code so that only the connection behaviour differs:

| mode | ops/sec | p99 |
|---|---|---|
| never reconnects | 56,104 | 6.59 ms |
| reconnects every 25 operations | 38,304 | 16.13 ms |
| reconnects every operation | 6,537 | 32.13 ms |

Roughly two dozen cache operations is what one PHP-FPM request tends to perform before its
connection closes, which puts a real site near the middle row: about a third of the throughput
lost and p99 roughly doubled, compared to never paying the connection cost. The gap widens as
concurrency rises rather than staying a flat per-connection tax.

## Environment

* Plugin version: 2.8.0
* PHP version: 8.x, PhpRedis
* WordPress version: 6.x

fx2300:
## Description

`build_parameters()` already builds a `persistent` connection parameter, but no constant feeds it
and only the PhpRedis cluster branch reads it, so it is `false` everywhere, always. This wires it
up:

- `persistent` joins the `$settings` list, so `WP_REDIS_PERSISTENT` is read like every other
  connection constant.
- PhpRedis and Relay call `pconnect()` when it is enabled. The fourth argument they already pass
  as `''` is `reserved` for `connect()` and `persistent_id` for `pconnect()`, so the identifier
  goes exactly where the existing call already had a slot.
- Credis receives the same identifier in its constructor's fourth argument, which is its
  persistence id.
- Predis needed no change at all: `connect_using_predis()` already passes `$parameters` to
  `new Predis\Client()`, which honours `persistent` — until now it was being handed an explicit
  `false`.
- A new `build_persistent_id()` returns `wp-db{database}` by default, overridable with
  `WP_REDIS_PERSISTENT_ID`.

Both constants are documented in the advanced options table, listed on the diagnostics screen,
and the diagnostics also report whether the connection ended up persistent.

## Motivation and Context

Fixes #632.

Why the identifier is derived from the database rather than fixed: persistent connections are
pooled per host, port and identifier, so state applied *after* connecting is part of the
connection, not of the request. `select()` is only called when the database is non-zero, so two
sites sharing one pool could otherwise silently inherit whichever database the other one selected
last — a data-correctness bug, not a performance detail. Including the database in the default
identifier keeps those pools separate.

`auth()` is re-issued on every request, so credentials need no equivalent treatment.

The measured cost of reconnecting per request, and the conditions it was measured under, are in
the issue.

## How Has This Been Tested?

On a staging site: WordPress 6.9, PHP 8.5 under PHP-FPM, PhpRedis 6.3.0, against a
Redis-protocol server on loopback (KeyDB, reporting 6.3.4), database 1, key prefix set.

`total_connections_received` from `INFO stats`, measured around bursts of 30 requests that fully
boot WordPress (a REST route, so no page cache is involved). Background drift with no traffic was
measured first, at about one connection per ten seconds, which accounts for the `+1` below:

| build | `WP_REDIS_PERSISTENT` | new connections per 30 requests |
| --- | --- | --- |
| stock 2.8.0 | unset | +31 |
| patched | unset | +31 |
| patched | `true` | +2 |

The middle row is the one that matters for anyone worried about regressions: with the constant
undefined, the patched build reconnects exactly as the stock one does.

Sampled during a burst, `CLIENT LIST` shows a single connection on **database 1** whose `age`
spans the whole burst rather than one connection per request, so `select()` is still being applied
to the pooled connection. `wp redis status` reports `Status: Connected`, `Drop-in: Valid`,
`Errors: []`, and the site's cache keys keep being written under their prefix in database 1.

Restoring the stock drop-in restores the old behaviour (10 requests, +11), so the effect is
attributable to the change and nothing else.

No latency difference was visible at this concurrency, and none was expected: that site's
PHP-FPM worker pool saturates well before the connection cost becomes visible to a visitor. The
gain this enables is measurable at the Redis protocol layer (see the issue) and matters under
concurrency, not on a sequential probe.

Static checks, PHP 8.5:

- `php -l` on both changed files: clean.
- `composer run phpstan`, with a relay-free bootstrap since `ext-relay` is not available locally:
  identical output before and after the change (same seven pre-existing errors, one line number
  shifted). No new finding.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

Off by default; behaviour is unchanged unless `WP_REDIS_PERSISTENT` is defined.

## Checklist:

- [x] My code follows the code style of this project
- [x] My change requires a change to the documentation
- [x] I have updated the documentation accordingly


fx2300:
Your workflows are `on: push`, so nothing runs on a pull request from a fork. I enabled Actions on
my fork and pushed the same commit there to get your suite to run against it. Reporting the result
here since the PR itself shows no checks.

`fx2300/redis-cache` at `b4aab23`, the exact commit this PR proposes -- 28 of 29 checks green:

| check | result |
| --- | --- |
| PHPStan (PHP 7.4 and 8.2, Relay 0.7.0) | pass |
| EditorConfig | pass |
| Testing, 15 Instance combinations (PHP 7.2-8.1, Redis 3.2.12 / 6.2.6 / 7.0.3) | pass |
| Testing, 10 Cluster combinations (PHP 7.2-8.1, Redis 6.2.8 / 7.0.7) | pass |
| Plugin Check | fail |

The Plugin Check failure is `readme.txt: Tested up to: 7.0 < 7.1`, and it is not related to this
change: the same job fails the same way on `develop` (run `33019724615`, alongside the same three
green jobs), and this PR does not touch `readme.txt`. I left it alone rather than bundling an
unrelated fix into this branch.

The cluster runs are the ones I would look at first, if you want a reason to trust this beyond the
diff: `persistent` was already read in the `RedisCluster` branch, where it could only ever be
`false`; with this change it can now be `true`, and that is the one path where existing behaviour
could have shifted rather than merely become reachable.

Runs, if you want to read them yourself:
https://github.com/fx2300/redis-cache/actions/runs/33204597098 (Code style)
https://github.com/fx2300/redis-cache/actions/runs/33204597047 (Run Tests)

includes/diagnostics.php
```diff
@@ -76,6 +77,8 @@
     'WP_REDIS_TIMEOUT',
     'WP_REDIS_READ_TIMEOUT',
     'WP_REDIS_RETRY_INTERVAL',
+    'WP_REDIS_PERSISTENT',
+    'WP_REDIS_PERSISTENT_ID',
```
tillkruss: Let's merge these. `WP_REDIS_PERSISTENT` can be `truthy` and we use the default `wp-db{database}`, or it can accept a `non-empty-string`.

tillkruss (changes requested):

includes/diagnostics.php
fx2300: Folded as requested: `WP_REDIS_PERSISTENT_ID` is gone, and `WP_REDIS_PERSISTENT` now carries both
answers -- any truthy value enables persistence with the default pool, a non-empty string also
names it.

`build_parameters()` assigns `constant( $constant )` with no cast, so the string reaches
`build_persistent_id()` intact. Exercised: `true` with database 0 gives `wp-db0`, `true` with
database 3 gives `wp-db3`, `'my-pool'` gives `my-pool`, `''` gives `wp-db3` and is falsy so it
stays on `connect()`.

The docblock now says why the default carries the database, and that a hand-given identifier is the
caller's to keep distinct: you pool by host, port and identifier, and `select()` only runs for a
non-zero database, so two sites sharing a pool would otherwise inherit whichever database the other
selected last.

Since a pull request from a fork runs none of your workflows, I ran them on my fork again at the
new commit. `fx2300/redis-cache` at `fa4c7b3`:

| check | result |
| --- | --- |
| Run Tests (25 jobs, real Redis: 15 Instance, 10 Cluster) | 25/25 pass |
| PHPStan (PHP 7.4 and 8.2, Relay 0.7.0) | pass |
| EditorConfig | pass |
| Plugin Check | fail, pre-existing |

Plugin Check fails on `readme.txt: Tested up to 7.0 < 7.1`, which this branch does not touch --
`readme.txt` is byte-for-byte `develop`. Same failure on `develop` itself (run 33019724615).

fx2300:
Thanks for merging, and for the review — confirmed on our side.

**Persistent connections: working.** Measured on staging against the merged change, 30 REST
requests that each boot WordPress: 31 new connections before, **2** after, with `CLIENT LIST`
showing a single connection on the database whose age covers the whole burst. With the constant
absent, behaviour is unchanged, so nothing moves for anyone not opting in.

We're running with it while waiting for a release to carry it, and will move to the official
version as soon as one ships.

Thank you for maintaining and developing this plugin — it carries our whole object cache and has
been solid throughout.

Best regards,
François

tillkruss:
This will be released shortly ✌️

PR #635, opened 2026-09-19
Respect `WP_REDIS_PREFIX` when flushing through the Predis helper

Pavlyshen:
## Problem

`Predis::flush()` in `includes/class-predis.php` ends in an unconditional `$this->redis->flushdb()`. It never looks at `WP_REDIS_PREFIX` or `WP_REDIS_SELECTIVE_FLUSH`, although `WP_Object_Cache::flush()` in the drop-in does.

Four paths reach that helper:

| Path | Call site |
|---|---|
| Plugin deactivation | `Plugin::on_deactivation()` |
| "Enable Object Cache" | `Plugin::do_admin_actions()`, `enable-cache` |
| "Disable Object Cache" | `Plugin::do_admin_actions()`, `disable-cache` |
| `wp redis enable` / `wp redis update-dropin` | `Commands::flush_redis()` |

So on an installation where several sites share one Redis database and are separated by `WP_REDIS_PREFIX`, the "Flush Cache" button correctly touches only that site's keys, while deactivating the plugin on one site empties the object cache of all of them. The prefix is the documented way to keep such sites apart, and on these paths it is silently ignored.

### Reproducing

Two WordPress installations pointed at the same Redis host and database:

```php
// site A
define( 'WP_REDIS_PREFIX', 'a:' );
define( 'WP_REDIS_SELECTIVE_FLUSH', true );

// site B
define( 'WP_REDIS_PREFIX', 'b:' );
define( 'WP_REDIS_SELECTIVE_FLUSH', true );
```

Warm both caches, then deactivate Redis Object Cache on site A. Site B loses its keys too, and so does anything else living in that database.

## The change

`Predis::flush()` now resolves the prefix the same way `wp_cache_init()` does (`WP_REDIS_PREFIX` constant, then the environment variable, then `WP_CACHE_KEY_SALT`, then the Cloudways fallback) and, when selective flushing is enabled, deletes only the keys carrying that prefix. Cluster connections keep iterating the masters, each one scanned instead of flushed. With no prefix, or with selective flushing off, the behaviour is exactly what it was.

One file, no new configuration, no change to the drop-in.

## Notes

- **No Lua here.** The drop-in runs its selective flush through `EVAL`, but this class holds no `redis_version` to decide whether the script needs a leading `redis.replicate_commands()` on Redis 3.2 to 4.x, and `EVAL` is unavailable on some managed hosts. A `SCAN` plus `DEL` loop is a handful of extra round trips on a path that runs once per activation or deactivation, and it does not block the server for the whole sweep.
- **`WP_REDIS_UNFLUSHABLE_GROUPS` is not honoured** by this method. Deleting everything under the prefix is what `FLUSHDB` already did, so this stays a strict narrowing of the current behaviour; teaching the helper about unflushable groups felt like a separate change.
- **Escaping is done with backslashes** rather than by reusing `WP_Object_Cache::glob_quote()`. That method wraps each special character in `[c]` through sequential `str_replace()` passes, so a bracket it inserted earlier is escaped again by a later pass. `glob_quote( '*:' )` returns `[[[]]*[]]:`, and `SCAN MATCH '[[[]]*[]]:*'` matches nothing, while `MATCH '\*:*'` and `MATCH '[*]:*'` both match `*:1`. It only bites prefixes containing one of `* + ? ! { } [ ] ( ) | @`, so I left the drop-in alone, but you may want to look at it separately.

## Verification

Tested against Redis 7.4 with Predis 2.4.1 on PHP 8.2.

Isolated runs on an empty database seeded with `a:1 a:2 a:3 ab:1 b:1 b:2 *:1 *:2 [x]:1 q?:1 noprefix`:

| Configuration | `flush()` | Keys deleted |
|---|---|---|
| prefix `a:`, selective on | `true` | `a:1 a:2 a:3` only, `ab:1` kept |
| prefix `b:` from the environment variable, selective on | `true` | `b:1 b:2` only |
| prefix `*:`, selective on | `true` | `*:1 *:2` only |
| prefix `[x]:`, selective on | `true` | `[x]:1` only |
| prefix `q?:`, selective on | `true` | `q?:1` only |
| prefix `a:`, selective off | falls through to `FLUSHDB` | unchanged path |
| no prefix, selective on | falls through to `FLUSHDB` | unchanged path |

End to end on a database shared by three installations, calling `( new Predis )->flush()` exactly as `on_deactivation()` does, from the site holding 1879 of the keys:

```
before   site 1: 1879   site 2: 6073   site 3: 7654
flush()  true
after    site 1:    2   site 2: 6073   site 3: 7654
```

Before the patch the same call took all 15606 keys.

Happy to add a CHANGELOG entry or reshape any of this if you would rather have it another way.


tillkruss:
That's intentional and `WP_REDIS_SELECTIVE_FLUSH` has been deprecated for years.

Pavlyshen:
Understood, thanks for the quick answer. Closing.

